Issue #78, opened 2019-06-24
env-web optimizations


Ivshti:
* [ ] fetch: optimize deserialization; curently we deserialize once in JS, then serialize to a string buffer to deserialize with serde on the WASM side; can be optimized
* [x] consider gzipping storage: https://github.com/Stremio/stremio-core/issues/82

Ivshti:
closing, there's no significant overhead to be worth fixing this

the fix would be:

- use `.array_buffer()` rather than `.json()`
- then cast: `js_sys::Uint8Array::new(&body)`
- then use `copy_to`

Ivshti:
So it was implemented in [env-web-optimization](https://github.com/Stremio/stremio-core/tree/env-web-optimization),

let's see the results:

### Firefox

#### with array_buffer

0-1ms, sometimes jumping to 4-5ms

#### with json

2-3, sometimes jumping to 5-7


### Chromium

Is more precise in it's measurement...

#### with array_buffer

parsing: 0.65380859375ms
parsing: 5.295166015625ms
 parsing: 5.617919921875ms
stremio_ng_example.js:66 parsing: 5.083251953125ms
stremio_ng_example.js:66 parsing: 5.64501953125ms
parsing: 0.617919921875ms
parsing: 2.054931640625ms
parsing: 0.634033203125ms
parsing: 0.06103515625ms
stremio_ng_example.js:66 parsing: 0.454833984375ms
stremio_ng_example.js:66 parsing: 0.052978515625ms
stremio_ng_example.js:66 parsing: 0.096923828125ms
stremio_ng_example.js:66 parsing: 0.042724609375ms
stremio_ng_example.js:66 parsing: 0.0390625ms


#### with json

parsing2: 3.597900390625ms
stremio_ng_example.js:60 parsing2: 5.43701171875ms
stremio_ng_example.js:60 parsing2: 8.896728515625ms
stremio_ng_example.js:60 parsing2: 7.054931640625ms
stremio_ng_example.js:60 parsing2: 5.672119140625ms
stremio_ng_example.js:60 parsing2: 0.05517578125ms
stremio_ng_example.js:60 parsing2: 0.72607421875ms
stremio_ng_example.js:60 parsing2: 0.6201171875ms
stremio_ng_example.js:60 parsing2: 3.265869140625ms
stremio_ng_example.js:60 parsing2: 1.56884765625ms
stremio_ng_example.js:60 parsing2: 0.656982421875ms
[...]

PR #85, opened 2019-07-28
Env web optimization

Ivshti:
the conclusion in #78 is probably inconclusive, cause

1) doesn't account for the overhead of the first `.json`
2) probably mostly measures time spent in `serde_json`

we should probably merge this


Ivshti:
premature optimization

PR #105, opened 2019-10-17
Default trait implemented for CatalogFiltered

NikolaBorislavovHristov:


NikolaBorislavovHristov:
i needed that default case in order to render initial ui for discover but now i realize this is wrong, because user could uninstall some of the default addons and for a short period of time the pickers will contain wrong data so i will close this

PR #106, opened 2019-10-18
catalog name fallbacks to catalog id

NikolaBorislavovHristov:


NikolaBorislavovHristov:
if some addon 'X' has 10 catalogs with ids 'top', 'rating', 'selected' you will see 10 entries with label 'X'
the catalog name must fallback to catalog id and in the ui we can show `{catalog_name} ({addon_name})` if we want to

PR #113, opened 2019-10-30
Meta item extend

NikolaBorislavovHristov:


src/state_types/models/notifications.rs
```diff
@@ -103,10 +103,12 @@ impl<Env: Environment + 'static> UpdateWithCtx<Ctx<Env>> for Notifications {
                                     // we have to get `now` somehow (environment or through a msg)
                                     // Alternatively, just set that when we add lib items
                                     .retain(|v| {
-                                        lib_item
-                                            .state
-                                            .last_vid_released
-                                            .map_or(false, |lvr| v.released > lvr)
+                                        v.released.map_or(false, |vr| {
```
Ivshti: 1) are you sure this is the same logic
2) released should be mandatory; if someone didn't provide it, then we'll default to 0 (1970)
3) this shows the benefit of rust - changing one small thing in the codebase will typecheck every usage; if this was JS or even TS we'd be fucked

src/types/meta_item.rs
```diff
@@ -54,8 +54,18 @@ pub struct MetaPreview {
     #[serde(default)]
     pub name: String,
     pub poster: Option<String>,
+    pub logo: Option<String>,
+    pub description: Option<String>,
+    pub release_info: Option<String>,
+    pub runtime: Option<String>,
+    pub released: Option<DateTime<Utc>>,
```
Ivshti: I'm not sure `released` and `release_info` should be there together
have you checked whether this is consistent with the SDK documentation?

also, please comment in the code that some fields are obsolete and should be removed, like `genres` and `directors`

src/types/meta_item.rs
```diff
@@ -100,13 +120,15 @@ pub struct Video {
     pub id: String,
     #[serde(alias = "name")]
     pub title: String,
-    pub released: DateTime<Utc>,
+    pub released: Option<DateTime<Utc>>,
```
Ivshti: I'm very much against this change, I said why earlier

src/types/meta_item.rs
NikolaBorislavovHristov: released is the date it was released.
release info is string like `2007-2009` or `2007-`

NikolaBorislavovHristov: genres and directors marked as deprecated

PR #119, opened 2019-11-01
Revert "DescriptorPreview dropped"

NikolaBorislavovHristov:
This reverts commit 560d723dee1d81fccb52462163571147e45bb85e.

### DIFF
--- a/src/types/addons/mod.rs
+++ b/src/types/addons/mod.rs
@@ -11,6 +11,13 @@ pub type TransportUrl = String;
 
 #[derive(PartialEq, Clone, Debug, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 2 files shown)


src/types/addons/mod.rs
```diff
@@ -9,6 +9,13 @@ use derive_more::*;
 
 pub type TransportUrl = String;
 
+#[derive(PartialEq, Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DescriptorPreview {
+    pub manifest: ManifestPreview,
```
Ivshti: it will be useful in the future to replace that with some light manifest, which only contains name/description/images, and not the full catalogs info, which may get heavy

PR #124, opened 2019-11-07
GDPRConsent added to Register

swetlasg:


src/state_types/models/context.rs
```diff
@@ -105,9 +106,15 @@ impl<Env: Environment + 'static> Update for Ctx<Env> {
                         None => Effects::msg(CtxUpdate(new_content).into()).unchanged(),
                     }
                 }
-                ActionUser::Register { email, password } => Effects::one(authenticate::<Env>(
+                ActionUser::Register { email, password, tos, privacy, marketing, from } => Effects::one(authenticate::<Env>(
                     action.to_owned(),
-                    APIRequest::Register { email, password },
+                    APIRequest::Register { email, password, gdpr_consent: GDPRConsent {
```
Ivshti: you can do tihs like JS: just specify `{ tos, privacy, ` etc.

src/state_types/msg/actions.rs
```diff
@@ -37,7 +37,7 @@ pub enum ActionSettings {
 #[serde(tag = "userOp", content = "args")]
 pub enum ActionUser {
     Login { email: String, password: String },
-    Register { email: String, password: String },
+    Register { email: String, password: String, tos: bool, privacy: bool, marketing: bool, from: String },
```
Ivshti: why is this not in a `gdpr_consent` field?

src/state_types/models/context.rs
Ivshti: if you implement my second comment, then this won't be needed

src/state_types/msg/actions.rs
NikolaBorislavovHristov: because we create the date in the core

NikolaBorislavovHristov: and the ui dont care about how data is structured

Ivshti: this is overkill; create the date in the UI

PR #131, opened 2019-11-11
Fix/crate name and travis

### LINKED ISSUE #125: proper crate name
currently, the derive macro fails when it's used in `stremio-core` itself

to solve this, we should check if the name of the currently compiled crate is `stremio-core`, and if so, use `"crate"`; else, use `crate_name("stremio-core").unwrap_or_else(||...panic...)`,

MartinKavik:
Fixes #125 

---

`.travis.yml` - I use pedantic clippy in my projects, example from Seed's repo:
```toml
command = "cargo"
args = ["clippy", "--all-features", "--", "--deny", "warnings", "--deny", "clippy::pedantic", "--deny", "clippy::nursery"]
```
=> can I add the similar command also into this project?
There is only `cargo clippy --all` in the current `.travis.yml` so clippy warnings are basically ignored.

---

And there are one clippy warning:
<details>
<summary>warning content</summary>

```
warning: large size difference between variants
  --> src\types\addons\mod.rs:65:5
   |
65 | /     Meta {
66 | |         // NOTE: we are not putting this in Option<>, since that way it gives us a valid
67 | |         // fallback to all of the previous variants, therefore resulting in inaccurate err messages
68 | |         // To support other /meta/ responses (meta extensions), we should make a MetaExt variant
69 | |         meta: MetaDetail,
70 | |     },
   | |_____^
   |
   = note: `#[warn(clippy::large_enum_variant)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_en
um_variant
help: consider boxing the large fields to reduce the total size of the enum
   |
69 |         meta: Box<MetaDetail>,
   |               ^^^^^^^^^^^^^^^

```
</details>

How do you want to resolve it?
- Add `Box`
- Decorate with `#[allow(clippy::...`
- Change size threshold

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+language: rust
+rust:
+  - stable
+
+os:
+  - linux
+
+jobs:
+    fast_finish: true
+
+cache: cargo
+
+env:
+  global:
+    - RUST_BACKTRACE=1
+
+install:
+  - rustup component add rustfmt
+  - rustup component add clippy
+
+script:
+  - cargo fmt --all -- --check
+  - cargo clippy --all
+  - cargo test --all
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,5 +20,5 @@ install:
 
 script:
   - cargo fmt --all -- --check
-  - cargo clippy --all
+  - cargo clippy --all -- -D warnings
   - cargo test --all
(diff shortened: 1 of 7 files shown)


Ivshti:
there's a `clippy.toml` currently which has increased the size threshold, but apparently now with the more stuff in `MetaDetail` this has happened again

I'd rather increase the overall limit again and once it breaks it, we'll box it

We are aware of the implications

PR #132, opened 2019-11-13
Add lib_items

swetlasg:


src/state_types/msg/actions.rs
```diff
@@ -60,6 +60,7 @@ pub enum ActionUser {
     PushAddons,
     LibSync,
     LibUpdate(LibItem),
+    LibItemsByType(String),
```
Ivshti: this action doesn't belong here; should be a Load action

src/state_types/models/context.rs
```diff
@@ -159,7 +159,9 @@ impl<Env: Environment + 'static> Update for Ctx<Env> {
                     None => Effects::none().unchanged(),
                 },
                 // We let the LibraryLoadable model handle this
-                ActionUser::LibSync | ActionUser::LibUpdate(_) => Effects::none().unchanged(),
+                ActionUser::LibSync | ActionUser::LibUpdate(_) | ActionUser::LibItemsByType(_) => {
```
Ivshti: this line will stay unmodified once `LibItemsByType` is removed from `ActionUser`

src/state_types/msg/actions.rs
Ivshti: `ActionUser` is specifically for user actions that are general but not related to particular models

src/state_types/models/lib_items.rs
```diff
@@ -0,0 +1,39 @@
+use super::{Ctx, LibraryLoadable};
+use crate::state_types::*;
+use crate::types::LibItem;
+use itertools::Itertools;
+use serde_derive::*;
+
+#[derive(Debug, Clone, Default, Serialize)]
+pub struct LibItems {
```
Ivshti: rename this to `LibraryFiltered` and the file to `library_filtered`

PR #157, opened 2020-07-02
Unit tests

swetlasg:


src/unit_tests/tests.rs
```diff
@@ -229,7 +238,16 @@ fn actionctx_login() {
             })),
         "profile updated successfully in storage"
     );
-    // TODO library updated successfully in storage
+    assert!(
+        STORAGE
+            .read()
+            .unwrap()
+            .get(LIBRARY_STORAGE_KEY)
+            .map_or(true, |data| {
```
nklhrstv, line 246: This is wrong. if there is no library in storage this must fail

src/unit_tests/tests.rs
```diff
@@ -98,32 +102,355 @@ fn actionctx_logout() {
...
+        "library updated successfully in memory"
+    );
+    assert!(
+        STORAGE
+            .read()
+            .unwrap()
+            .get(PROFILE_STORAGE_KEY)
+            .map_or(true, |data| {
+                serde_json::from_str::<Profile>(&data).unwrap().auth
+            } == Some(Auth {
```
nklhrstv, line 228: this syntax in super strange.
can you cahnge it to
```
.map_or(true, |data| {
    serde_json.... == Some()
})
```

PR #162, opened 2020-07-13
Uninstall addon tests

swetlasg:


src/unit_tests/tests.rs
```diff
@@ -721,3 +721,345 @@ fn actionctx_installaddon_update_fail() {
...
+                    ..Default::default()
+                },
+                ..Default::default()
+            },
+        },
+        1000,
+    );
+    run(
+        runtime.dispatch(&Msg::Action(Action::Ctx(ActionCtx::UninstallAddon(
+            addon.to_owned().transport_url,
```
nklhrstv, line 971: that to_owned call is in a strange place. it usually put last to avoid unnecessary mem copy.
you could use hardcoded strings here

src/unit_tests/tests.rs
```diff
@@ -721,3 +721,345 @@ fn actionctx_installaddon_update_fail() {
...
+    );
+    assert_eq!(
+        serde_json::from_str::<Profile>(&STORAGE.read().unwrap().get(PROFILE_STORAGE_KEY).unwrap())
+            .unwrap()
+            .addons
+            .len(),
+        1,
+        "There is one addon in storage"
+    );
+    assert_eq!(
```
nklhrstv, line 992: can you combine those two asserts in one like
assert_eq!(xxxx.addons, vec![addon.to_owned()])

src/unit_tests/tests.rs
```diff
@@ -721,3 +721,345 @@ fn actionctx_installaddon_update_fail() {
...
+            },
+        },
+        1000,
+    );
+    run(
+        runtime.dispatch(&Msg::Action(Action::Ctx(ActionCtx::UninstallAddon(
+            "transport_url2".to_owned(),
+        )))),
+    );
+    assert_eq!(
```
nklhrstv, line 1052: assert that .addons contains `addon`

nklhrstv:
installaddon tests could be named like _update, _install, _already_installed

src/unit_tests/tests.rs
```diff
@@ -721,3 +721,327 @@ fn actionctx_installaddon_update_fail() {
...
+            .app
+            .read()
+            .unwrap()
+            .ctx
+            .profile
+            .addons
+            .contains(&addon),
+        "addon is in memory"
+    );
+    assert!(
```
nklhrstv, line 1040: use assert!(STORAGE...map_or(false, )) code structure here to be consistent with other tests

src/unit_tests/tests.rs
```diff
@@ -721,3 +721,327 @@ fn actionctx_installaddon_update_fail() {
...
+        runtime
+            .app
+            .read()
+            .unwrap()
+            .ctx
+            .profile
+            .addons
+            .contains(&addon),
+        "addon is in memory"
```
nklhrstv, line 1038: the message should be `addons in memory not updated`

src/unit_tests/tests.rs
```diff
@@ -721,3 +721,327 @@ fn actionctx_installaddon_update_fail() {
...
+    );
+    assert!(
+        serde_json::from_str::<Profile>(&STORAGE.read().unwrap().get(PROFILE_STORAGE_KEY).unwrap())
+            .unwrap()
+            .addons
+            .contains(&addon),
+        "addon is in storage"
```
nklhrstv, line 1045: the message should be `addons in strorage not updated`

src/unit_tests/tests.rs
```diff
@@ -721,3 +721,327 @@ fn actionctx_installaddon_update_fail() {
...
+    );
+}
+
+#[test]
+fn actionctx_uninstalladdon_not_installed() {
+    #[derive(Model, Debug, Default)]
+    struct Model {
+        ctx: Ctx<Env>,
+    }
+    let addon = Descriptor {
```
nklhrstv, line 984: you need only `let profile = ...` here

PR #163, opened 2020-07-16
Install addon tests

swetlasg:


src/state_types/models/ctx/update_profile.rs
```diff
@@ -96,6 +96,14 @@ pub fn update_profile<Env: Environment + 'static>(
                     .iter()
                     .map(|addon| &addon.transport_url)
                     .position(|transport_url| *transport_url == addon.transport_url);
+                let push_to_api_effects = match &profile.auth {
```
nklhrstv, line 99: that block needs to be moved below the next block because that way an older version of profile.addons will be pushed to API

src/unit_tests/tests.rs
```diff
@@ -534,17 +534,127 @@ fn actionctx_installaddon_install() {
...
+        REQUESTS.read().unwrap().len(),
+        1,
+        "One request has been sent"
+    );
+    assert_eq!(
+        REQUESTS.read().unwrap().get(0).unwrap().to_owned(),
+        Request {
+            url: "https://api.strem.io/api/addonCollectionSet".to_owned(),
+            method: "POST".to_owned(),
+            body: "{\"type\":\"AddonCollectionSet\",\"authKey\":\"auth_key\",\"addons\":[]}"
```
nklhrstv, line 652: thats not possible!!! addons vec cannot be empty after installing an addon

PR #171, opened 2020-07-29
Add to library test

swetlasg:


src/unit_tests/ctx/add_to_library.rs
```diff
@@ -143,3 +144,137 @@ fn actionctx_addtolibrary() {
...
+                            fb_id: None,
+                            avatar: None,
+                            last_modified: Env::now(),
+                            date_registered: Env::now(),
+                        },
+                    }),
+                    ..Default::default()
+                },
+                library: LibBucket {
+                    uid: None,
```
nklhrstv, line 219: uid cannot be None when there is a user

src/unit_tests/ctx/add_to_library.rs
```diff
@@ -143,3 +144,137 @@ fn actionctx_addtolibrary() {
...
+                    items: items,
```
nklhrstv, line 220: inline items hash map here

src/unit_tests/ctx/add_to_library.rs
```diff
@@ -143,3 +144,137 @@ fn actionctx_addtolibrary() {
...
+        released: None,
+        poster_shape: Default::default(),
+        trailer: None,
+        behavior_hints: Default::default(),
+    };
+    let lib_item = LibItem {
+        id: "id".to_owned(),
+        removed: false,
+        temp: false,
+        ctime: Some(Utc.ymd(2020, 1, 1).and_hms_milli(0, 0, 0, 0)),
```
nklhrstv, line 187: the concept of this test is to prove that ctime and state wont be updated but mtime will be.

src/unit_tests/ctx/add_to_library.rs
```diff
@@ -143,3 +144,137 @@ fn actionctx_addtolibrary() {
         "datastorePut request has been sent"
     );
 }
+
+#[test]
+fn actionctx_addtolibrary_already_added() {
```
nklhrstv, line 149: you can do this test without a user

src/unit_tests/ctx/add_to_library.rs
```diff
@@ -0,0 +1,297 @@
...
+        released: None,
+        poster_shape: Default::default(),
+        trailer: None,
+        behavior_hints: Default::default(),
+    };
+    let lib_item = LibItem {
+        id: "id".to_owned(),
+        removed: false,
+        temp: false,
+        ctime: Some(Env::now()),
```
nklhrstv, line 171: dont use Env in tests before reset. Use custom time here.
check in previous tests for that mistake.

src/unit_tests/ctx/add_to_library.rs
```diff
@@ -0,0 +1,297 @@
...
+        mtime: Env::now(),
+        state: Default::default(),
```
nklhrstv, line 173: use some none default state here so you can really confirm that its updated properly

src/unit_tests/ctx/add_to_library.rs
```diff
@@ -0,0 +1,297 @@
...
+        poster: None,
+        poster_shape: Default::default(),
+        behavior_hints: Default::default(),
+    };
+    Env::reset();
+    let (runtime, _) = Runtime::<Env, Model>::new(
+        Model {
+            ctx: Ctx {
+                library: LibBucket {
+                    uid: Some("id".to_owned()),
```
nklhrstv, line 185: having Some() uid in libbucket and no auth is wrong. they must be in sync

src/unit_tests/ctx/add_to_library.rs
```diff
@@ -0,0 +1,297 @@
...
+            .app
+            .read()
+            .unwrap()
+            .ctx
+            .library
+            .items
+            .get(&meta_preview.id)
+            .unwrap()
+            .mtime
+            > Some(&lib_item).unwrap().mtime,
```
nklhrstv, line 244: asset that mtime is Env::now().

src/unit_tests/ctx/add_to_library.rs
```diff
@@ -0,0 +1,297 @@
...
+    );
+}
+
+#[test]
+fn actionctx_addtolibrary_already_added() {
+    #[derive(Model, Debug, Default)]
+    struct Model {
+        ctx: Ctx<Env>,
+    }
+    let meta_preview = MetaPreview {
```
nklhrstv, line 153: use some none default prop here so you can confirm it really updated

src/unit_tests/ctx/add_to_library.rs
```diff
@@ -0,0 +1,297 @@
...
+    );
+    assert!(
+        STORAGE
+            .read()
+            .unwrap()
+            .get(LIBRARY_RECENT_STORAGE_KEY)
+            .map_or(false, |data| {
+                serde_json::from_str::<(UID, Vec<LibItem>)>(&data)
+                    .unwrap()
+                    .1[0]
```
nklhrstv, line 255: this dont look nice. change it.

src/unit_tests/ctx/add_to_library.rs
```diff
@@ -0,0 +1,297 @@
...
+            .app
+            .read()
+            .unwrap()
+            .ctx
+            .library
+            .items
+            .get(&meta_preview.id)
+            .unwrap()
+            .ctime,
+        Some(&lib_item).unwrap().ctime,
```
nklhrstv, line 216: wrapping a reference in Some and then unwrap it is the worst way of cloning something. seems like youve used some of the previous asserts and just made it to work.
could it be just lib_item.ctime

src/unit_tests/ctx/add_to_library.rs
```diff
@@ -0,0 +1,297 @@
...
+                    .state
+                    == Some(&lib_item).unwrap().state
+            }),
+        "recent library item state not changed in storage"
+    );
+    assert!(
+        STORAGE
+            .read()
+            .unwrap()
+            .get(LIBRARY_RECENT_STORAGE_KEY)
```
nklhrstv, line 279: deserializing the same thing 3 times in one test is unnecessary

PR #178, opened 2020-08-14
Update README.md

ozkanpakdil:
dead links cleared, feel free to update them if they exist somewhere else. I was looking for to write a small change on player btw :)


Ivshti:
not dead links, just links to private repos

PR #181, opened 2020-08-20
Sync library with api test

swetlasg:


src/unit_tests/ctx/sync_library_with_api.rs
```diff
@@ -295,3 +295,251 @@ fn actionctx_synclibrarywithapi_with_user() {
...
+            }),
+        "Library recent slot updated successfully in storage"
+    );
+    assert_eq!(
+        REQUESTS.read().unwrap().len(),
+        3,
+        "Three requests have been sent"
+    );
+    assert_eq!(
+        REQUESTS.read().unwrap().get(0).unwrap().to_owned(),
```
nklhrstv, line 519: no need to check for the body here, only for the url

src/unit_tests/ctx/sync_library_with_api.rs
```diff
@@ -295,3 +295,251 @@ fn actionctx_synclibrarywithapi_with_user() {
...
+        },
+        "datastoreMeta request has been sent"
+    );
+    assert_eq!(
+        REQUESTS.read().unwrap().get(1).unwrap().url,
+        "https://api.strem.io/api/datastorePut".to_owned(),
+        "datastorePut request has been sent"
+    );
+    assert_eq!(
+        REQUESTS.read().unwrap().get(2).unwrap().to_owned(),
```
nklhrstv, line 534: this one is unnecessary. havent thought about that while writing the sync logic.
if there are no items to be pulled/pushed no requests should be sent.

nklhrstv (changes requested):

src/state_types/models/ctx/update_library.rs
```diff
@@ -169,20 +169,34 @@ pub fn update_library<Env: Environment + 'static>(
             result,
         )) if Some(loading_auth_key) == auth_key => match result {
             Ok((pull_ids, push_ids)) => {
-                let push_items = library
+                let push_items: Vec<_> = library
```
nklhrstv, line 172: use `collect::<Vec<_>>` instead of this

nklhrstv (changes requested):

src/state_types/models/ctx/update_library.rs
```diff
@@ -175,14 +175,25 @@ pub fn update_library<Env: Environment + 'static>(
                     .filter(move |(id, _)| push_ids.iter().any(|push_id| push_id == *id))
                     .map(|(_, item)| item)
                     .cloned()
-                    .collect();
+                    .collect::<Vec<_>>();
+                let push_items_to_api_effects = if push_items.is_empty() {
+                    Effects::none().unchanged()
+                } else {
+                    Effects::one(push_items_to_api::<Env>(push_items, loading_auth_key))
```
nklhrstv, line 182: you need to add .unchanged here also

nklhrstv (changes requested):
Do you need to use .to_owned in Event::LibrarySyncWithAPIPlanned anymore?

PR #190, opened 2020-10-13
MetaItem serializing/deserializing test 

swetlasg:


src/unit_tests/serde/poster_shape.rs
```diff
@@ -0,0 +1,42 @@
+use crate::types::resource::PosterShape;
+use serde_test::{assert_de_tokens, assert_tokens, Token};
+
+#[test]
+fn ser_de_poster_shape() {
```
nklhrstv, line 5: name those functions with the name of the struct only

nklhrstv (approved):

PR #193, opened 2020-10-16
DescriptorFlags serializing/deserializing test

swetlasg:


src/unit_tests/serde/descriptor_flags.rs
```diff
@@ -0,0 +1,36 @@
...
+    );
+    assert_de_tokens(
+        &DescriptorFlags {
+            official: false,
+            protected: false,
+        },
+        &[
+            Token::Struct {
+                name: "DescriptorFlags",
+                len: 2,
```
nklhrstv, line 31: should be `len: 0`

nklhrstv (changes requested):

nklhrstv (approved):

PR #207, opened 2020-10-28
ApiRequest serializing test

swetlasg:


src/types/api/request.rs
```diff
@@ -41,14 +44,13 @@ impl APIMethodName for APIRequest {
     }
 }
 
-#[derive(Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Derivative, Clone, PartialEq, Serialize, Deserialize)]
+#[derivative(Default)]
```
nklhrstv, line 48: this should be derived only in test cfg

nklhrstv (changes requested):

src/types/api/request.rs
```diff
@@ -41,14 +44,13 @@ impl APIMethodName for APIRequest {
     }
 }
 
-#[derive(Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Derivative, Clone, PartialEq, Serialize, Deserialize)]
```
nklhrstv, line 47: `Derivative` should be derived only in tests

nklhrstv (changes requested):

nklhrstv (approved):

PR #215, opened 2020-11-06
StreamSource serializing/deserializing test

swetlasg:


nklhrstv (changes requested):
there is duplicated code that could be refactored

nklhrstv (approved):

PR #250, opened 2021-06-20
Add seek time properties to settings

tymmesyde:
Related to https://github.com/Stremio/stremio-web/issues/145

Cause a serialization error when localStorage.profile.settings has already been set but can be dismissed by clearing the localStorage.
I don't think this is a suited behavior but I don't really understand why this is happening, i believe it comes from `runtime/env.rs`


nklhrstv:
the crash is because a migration is required [here](https://github.com/Stremio/stremio-core/blob/development/src/runtime/env.rs#L129)

nklhrstv (approved):

PR #279, opened 2022-03-07
Handle imdb_rating and genres on MetaItem

tymmesyde:


nklhtv:
Those should be used from .links api or deserialized as links using the serde custom deserializer

Issue #323, opened 2022-06-10
Improve imdbRating shimming


TheBeastLT:
Currently legacy `imdbRating` property is specified as a string, but some addons return it as a number in the JSON. Then catalog/meta responses from such addons can no longer be parsed.
Also it fails parsing on null shimmed properties - `genres: null` / `trailers: null`

PR #327, opened 2022-06-15
Better MetaItem deserialization

core1024:
poster, background and logo as Url type. Parse correctly genres and trailers if null. Accepts imdbRating as floating point number.

Fixes https://github.com/Stremio/stremio-core/issues/323


src/types/resource/meta_item.rs
```diff
@@ -34,20 +35,32 @@ struct MetaItemPreviewLegacy {
...
     release_info: Option<String>,
     runtime: Option<String>,
     released: Option<DateTime<Utc>>,
     #[serde(default)]
     poster_shape: PosterShape,
+    #[serde(
+        default,
+        deserialize_with = "deserialize_number_or_text_to_some_string"
+    )]
     imdb_rating: Option<String>,
```
nklhtv, line 57: can you try to use something like `#[serde_as(deserialize_as = "Option<NumberOrString>")]`
where `NumberOrString` is a struct that implemented [DeserializeAs](https://docs.rs/serde_with/latest/serde_with/trait.DeserializeAs.html) similar to [this](https://github.com/Stremio/stremio-core/blob/catalog-pagination/src/types/serde_as_ext.rs#L14-L31)

nklhtv (changes requested):

nklhtv (approved):

PR #337, opened 2022-07-07
Implement and add streams bucket

### LINKED ISSUE #65: Player model: save stream ID (hash)
save the hash (or prefix of hash) of the last playing stream in the `LibItem`, in order to be able to identify it on another device as well

TheBeastLT:
Closes #65 

Just not sure if we should overwrite and erase currently existing `StreamsBucket` when a new login is successful, since these are not persisted in the API like `LibraryBucket`, so we'd loose this info. Maybe it would be better to keep the `StreamsBucket` global for the installation, regardless of the user?


micwoj92:
This branch has conflicts that must be resolved

unclekingpin:
Closed as its part if #466

PR #354, opened 2022-07-24
add new secondary language settings

TheBeastLT:


src/types/profile/settings.rs
```diff
@@ -38,7 +40,9 @@ impl Default for Settings {
             streaming_server_url: STREAMING_SERVER_URL.to_owned(),
             interface_language: "eng".to_owned(),
             audio_language: "eng".to_owned(),
+            secondary_audio_language: "".to_owned(),
```
nklhtv, line 43: if those are optional lets use Option<String>

nklhtv (changes requested):

src/types/profile/settings.rs
TheBeastLT: I guess then we don't need a migration?

nklhtv: seems like it, will test it

nklhtv (approved):

PR #369, opened 2022-08-19
Trakt integration

nklhtv:


src/models/player.rs
```diff
@@ -120,6 +183,14 @@ impl<E: Env + 'static> UpdateWithCtx<E> for Player {
                     .join(watched_effects)
             }
             Msg::Action(Action::Unload) => {
+                let ended_effects = if !self.ended && self.selected.is_some() {
```
TheBeastLT, line 186: Maybe we should have similar logic as in `switch_to_next_video_effects` where we check if it was watched after certain threshold, so we could mark it as ended, so it doesn't hang in trakt?

TheBeastLT (approved):

PR #390, opened 2022-10-31
handle streaming server casting route

tymmesyde:
Handle devices response from streaming server `/casting` route
Allow to use the `/casting/{device}/player` route with a new `PlayOnDevice` action for `StreamingServer` model


elpiel (commented):
Code LGTM. Have you tested this?

src/models/streaming_server.rs
```diff
@@ -445,3 +514,32 @@ fn parse_torrent(torrent: &[u8]) -> Result<(String, Vec<String>), serde_bencode:
...
+
+fn play_on_device<E: Env + 'static>(url: &Url, args: &PlayOnDeviceArgs) -> Effect {
+    #[derive(Serialize)]
+    #[serde(rename_all = "camelCase")]
+    struct Body {
+        source: String,
+        time: u64,
+    }
+    let endpoint = url
+        .join(&format!("casting/{}/player", args.device))
```
elpiel: Could you give an example of the device argument? We should be certain that it does not need `urlencode` or always `urlencode` (which would be easier)

tymmesyde: Sure, you can check this action dispatch on stremio-web:
https://github.com/Stremio/stremio-web/blob/development/src/routes/Player/OptionsMenu/OptionsMenu.js#L52
It's just a string that represent an external 'casting device' id returned by the server `/casting` route

PR #391, opened 2022-11-01
Additional new settings

TheBeastLT:


src/types/profile/settings.rs
```diff
@@ -9,10 +9,13 @@ use url::Url;
 pub struct Settings {
     pub interface_language: String,
     pub streaming_server_url: Url,
+    pub player_type: u8,
```
TheBeastLT: Not sure what's the best type to use for this, so that it's still app agnostic. It could be an int, enum or a string. For Android TV I'll have these values `0=Exo/1=libVLC/2=External`

tymmesyde: Could we separate those settings like `internal_player` (Default/Exo/libVLC) and `external_player` (None/mpv/vlc, ...)
It will probably needed for a feature on iOS to always start the stream in external player
Not sure 100% about that if it will be needed

elpiel: Enum and string seem like a good option, although I don't fully understand the domain yet.
What are possible values for the players and how often/how much they need to change?

nklhtv: i would suggest a string.
thats how we handle other settings like language, colors and font

TheBeastLT: > Could we separate those settings like internal_player (Default/Exo/libVLC) and external_player (None/mpv/vlc, ...)

@tymmesyde Not sure why it couldn't be a single option and include both internal player(s) and the external ones?

> i would suggest a string.

Yeah, makes sense then to use it here as well.

tymmesyde: Just thought that if a user is setting it to Exo on his tv, it would conflict on iOS or desktop if the user chose to use external player but it was just a suggestion

TheBeastLT: @tymmesyde settings are local per platform, they are not synced, so it can have different values per app/platform.

src/runtime/env.rs
```diff
@@ -368,3 +374,34 @@ fn migrate_storage_schema_to_v5<E: Env>() -> TryEnvFuture<()> {
...
+            match profile
+                .as_mut()
+                .and_then(|profile| profile.as_object_mut())
+                .and_then(|profile| profile.get_mut("settings"))
+                .and_then(|settings| settings.as_object_mut())
+            {
+                Some(settings) => {
+                    settings.insert(
+                        "player_type".to_owned(),
+                        serde_json::Value::Number(serde_json::Number::from(0)),
```
nklhtv: player_type is declared as Option<String>
but here we put 0
I suggest we use serde_json::Value::NULL

nklhtv (changes requested):

src/runtime/env.rs
TheBeastLT: Actually, we can just remove it, since it should deserialize to `None` if there's nothing there (same was done with secondary languages settings). Tho maybe it would be good to check whether the `play_in_external_player` flag is set and store `external` in the `player_type`, tho not entirely sure how to impl this.

nklhtv: i thought we should use player_type only for the internal player preference
if we use it for external too maybe we can combine those 2 settings

TheBeastLT: Yeah, I think it makes sense to just have one setting, since you can only use one player, so `external` is also a sort of player in that sense.

nklhtv: yes, so the migration should also delete `play_in_external` bool and insert `external` if that flag was previously set to true.
merging of this pr will then require changes in the apps.
we can keep the checkbox for 'play in external player' but instead of setting the flag to true we can set that string to `external`

TheBeastLT: Yes, makes sense to remove it. I've added a migration for it, but would be good if you could check it if everything's correct.

src/runtime/env.rs
```diff
@@ -368,3 +374,39 @@ fn migrate_storage_schema_to_v5<E: Env>() -> TryEnvFuture<()> {
...
+                .as_mut()
+                .and_then(|profile| profile.as_object_mut())
+                .and_then(|profile| profile.get_mut("settings"))
+                .and_then(|settings| settings.as_object_mut())
+                .map(|settings| (settings.remove("playInExternalPlayer"), settings))
+            {
+                Some((Some(play_in_external_player), settings)) => {
+                    settings.insert(
+                        "playerType".to_owned(),
+                        if play_in_external_player == "true" {
```
elpiel: I'm writing tests about the migration but I wanted to point out that this is incorrect just in case someone wants to merge the PR before I finish with the tests.

A `Value::Bool` cannot be compared with a str of `"true"`

elpiel (changes requested):

nklhtv (approved):

PR #392, opened 2022-11-08
install trakt addon

nklhtv:


elpiel (approved):
LGTM
PS: Thanks for walking me through

PR #398, opened 2022-11-24
Additional settings tests

elpiel:
I needed access to TestEnv outside of `unit_tests` so I merged `development`

### DIFF
--- a/src/runtime/env.rs
+++ b/src/runtime/env.rs
@@ -383,17 +383,15 @@ fn migrate_storage_schema_to_v6<E: Env>() -> TryEnvFuture<()> {
                 .and_then(|profile| profile.as_object_mut())
                 .and_then(|profile| profile.get_mut("settings"))
                 .and_then(|settings| settings.as_object_mut())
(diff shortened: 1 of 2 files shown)


src/runtime/env.rs
```diff
@@ -388,7 +388,7 @@ fn migrate_storage_schema_to_v6<E: Env>() -> TryEnvFuture<()> {
                 Some((Some(play_in_external_player), settings)) => {
                     settings.insert(
                         "playerType".to_owned(),
-                        if play_in_external_player == serde_json::Value::Bool(true) {
```
elpiel, line 391: @TheBeastLT you can also compare `serde_json::Value` directly to `bool`
https://docs.rs/serde_json/latest/serde_json/enum.Value.html#impl-PartialEq%3Cbool%3E-for-Value
I already had this change made (fixing the bug) that's why I left it in this PR

TheBeastLT: Ahh nice, thanks! My IDE didn't resolve the type, so I didn't explore too much what is available.

PR #405, opened 2022-12-05
change defaults for some settings

TheBeastLT:
After some discussion with @jaruba some defaults are to be changed:
 - Enabled binge watching by default (same as in desktop app)
 - Enable next video notification at 35000ms (same as in desktop app)
 - Change seek duration to 10000ms, to allow seeking more precisely and it's the default in some other streaming platforms
 - Change outline color to black, use the 6 symbol hex code without alpha, as web and android have different 8hex definitions (rgba vs argb)


elpiel (commented):
- `nextVideoNotificationDuration` - for this setting, have we deployed the migration yet? If not, I think it's good to change it in the migration, otherwise - not.

- regarding the other options - these will not take effect to current users, as we don't use a migration for them and the old values will stay. Is this the intended consequence?

TheBeastLT:
> * `nextVideoNotificationDuration` - for this setting, have we deployed the migration yet? If not, I think it's good to change it in the migration, otherwise - not.

Yes, correct it was not deployed anywhere yet.

>     * regarding the other options - these will not take effect to current users, as we don't use a migration for them and the old values will stay. Is this the intended consequence?

Yes, correct these will only affect new installs or when a user relogs (as that clears the settings).

jaruba:
@elpiel we discussed these defaults changes y-day, and we think they fit better for new users, we should obviously not overwrite the user settings of current users though

elpiel:
> @elpiel we discussed these defaults changes y-day, and we think they fit better for new users, we should obviously not overwrite the user settings of current users though

Maybe we can check if they are default and if so - change them?
I've never altered my profile settings but I would like to get these changes out-of-the-box if they are making my UX with the app better.

jaruba:
@elpiel way too complex, we should not try to detect defaults, or change current user settings

elpiel:
> @elpiel way too complex, we should not try to detect defaults, or change current user settings

It's just a suggestion and we need to do a migration anyway :shrug: 

- The only difference with performing the changes to all users with default settings would be to check if the current value (if one exist) is the old default value.
- This is more relevant to other settings and not `subtitlesOutlineColor` from what I can see.
- `nextVideoNotificationDuration` was introduced in v6 migration which is the latest and it was `0` so I think we should better change it

@TheBeastLT Shouldn't we also change the other subtitles color settings as they also use 8hex definition? I.e.:
```rust
subtitles_text_color: "#FFFFFFFF".to_owned(),
subtitles_background_color: "#00000000".to_owned(),
```

jaruba:
> The only difference with performing the changes to all users with default settings would be to check if the current value (if one exist) is the old default value.

This is considered bad practice though. The people that used the app for months now got used to it working in a certain way. For example: pressing right 4x times to skip an intro, yet we would change the skip duration now for them too and they would need to press 8x times, this would break UX for them.

They always have the option of changing these settings themselves, but we should not change it for them, only for new users.

TheBeastLT:
>     * `nextVideoNotificationDuration` was introduced in v6 migration which is the latest and it was `0` so I think we should better change it

This migration did not run anywhere yet, so it will insert the correct default.

> @TheBeastLT Shouldn't we also change the other subtitles color settings as they also use 8hex definition? I.e.:
> 
> ```rust
> subtitles_text_color: "#FFFFFFFF".to_owned(),
> subtitles_background_color: "#00000000".to_owned(),
> ```

We could change text color to 6hex, but not background as now it's transparent and in 6hex it would become black. IMO it's fine as is. Maybe @nklhtv could offer his opinion as well.

src/unit_tests/data_export.rs
```diff
@@ -1,11 +1,10 @@
 use crate::models::common::Loadable;
 use crate::models::ctx::Ctx;
 use crate::models::data_export::DataExport;
-use crate::runtime::msg::{Action, ActionCtx, ActionLoad};
```
elpiel: could you revert these changes please, I'm fixing this in #403 and will cause conflicts

elpiel (approved):
I don't know what is the reasoning of Niki for making a new migration even when this has not been deployed to users yet.

I'm approving it but we need to consult Niki on this question before merging

PR #411, opened 2022-12-22
fix(data_export): use api subdomain

tymmesyde:


elpiel (approved):
I don't know where I copied the URL from to make this mistake...
Good catch!

PR #414, opened 2023-01-03
Add Video Params for Subs Requests

jaruba:
Related PRs:
- https://github.com/Stremio/stremio-video/pull/39
- https://github.com/Stremio/stremio-web/pull/336

### DIFF
--- a/src/models/player.rs
+++ b/src/models/player.rs
@@ -1,4 +1,6 @@
-use crate::constants::{CREDITS_THRESHOLD_COEF, WATCHED_THRESHOLD_COEF};
+use crate::constants::{
+    CREDITS_THRESHOLD_COEF, VIDEO_HASH_EXTRA_PROP, VIDEO_SIZE_EXTRA_PROP, WATCHED_THRESHOLD_COEF,
+};
 use crate::models::common::{
     eq_update, resource_update, resources_update_with_vector_content, Loadable, ResourceAction,
     ResourceLoadable, ResourcesAction,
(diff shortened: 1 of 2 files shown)


src/models/player.rs
```diff
@@ -42,13 +44,21 @@ pub struct AnalyticsContext {
     pub has_trakt: bool,
 }
 
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct VideoParams {
+    pub hash: Option<String>,
+    pub size: Option<u32>,
```
elpiel, line 50: wouldn't it better to stick with the same field names as the extra properties?

src/models/player.rs
```diff
@@ -1,20 +1,22 @@
...
+use crate::types::addon::{AggrRequest, ExtraExt, ResourcePath, ResourceRequest};
 use crate::types::library::{LibraryBucket, LibraryItem};
 use crate::types::profile::Settings as ProfileSettings;
 use crate::types::resource::{MetaItem, SeriesInfo, Stream, Subtitles, Video};
 use chrono::{DateTime, Utc};
 use itertools::Itertools;
 use serde::{Deserialize, Serialize};
-use std::cmp;
 use std::marker::PhantomData;
+use std::{cmp, str};
```
elpiel: `str` is not required here, not sure why the linters do not complain about it though.

elpiel (approved):

src/models/player.rs
```diff
@@ -119,14 +128,29 @@ impl<E: Env + 'static> UpdateWithCtx<E> for Player {
...
-                let subtitles_effects = match &selected.subtitles_path {
-                    Some(subtitles_path) => resources_update_with_vector_content::<E, _>(
-                        &mut self.subtitles,
-                        ResourcesAction::ResourcesRequested {
-                            request: &AggrRequest::AllOfResource(subtitles_path.to_owned()),
-                            addons: &ctx.profile.addons,
-                        },
-                    ),
+                let subtitles_effects = match (&selected.subtitles_path, &selected.video_params) {
+                    (Some(subtitles_path), Some(video_params)) => {
```
TheBeastLT, line 132: @jaruba We should send the subtitles requests even if `video_params` is `None`?

TheBeastLT:
@jaruba Also the legacy transport should be updated to include these new props - https://github.com/Stremio/stremio-core/blob/development/src/addon_transport/http_transport/legacy/mod.rs#L204
Query should look like `"query":{"itemHash":"tt11570202 1 1","videoHash":"f27ea3166eff6f33","videoSize":358912372}}`

PR #421, opened 2023-01-13
feat(deeplinks): return download streaming server url for torrent streams

tymmesyde:


src/types/resource/stream.rs
```diff
@@ -92,6 +96,82 @@ impl Stream {
...
+                    file_idx,
+                    announce,
+                },
+                Some(streaming_server_url),
+            ) => {
+                let mut url = streaming_server_url.to_owned();
+                match url.path_segments_mut() {
+                    Ok(mut path) => {
+                        path.push(&hex::encode(info_hash));
+                        if let Some(file_idx) = file_idx {
```
TheBeastLT, line 138: @tymmesyde If I'm reading correctly, if `file_index` is `None` then the created url won't contain the file index path and will be just `/infoHash`? This is not a supported path by enginefs as far as I know, since you always need to provide and an index (tho it would be great if you could provide smth like `-1` for index, then you could construct such url statically always), correct  @jaruba ? 
So if `file_index` is `None` you can't really construct a streaming url, as first you need to do a POST request to enginefs to resolve the largest file and get it's index, which you can then use to construct such GET url.

PR #423, opened 2023-01-17
feat: Notifications with Continue watching

elpiel:


src/constants.rs
```diff
@@ -27,6 +29,7 @@ pub const GENRES_LINK_CATEGORY: &str = "Genres";
 pub const CINEMETA_TOP_CATALOG_ID: &str = "top";
 pub const IMDB_TITLE_PATH: &str = "title";
 pub const YOUTUBE_ADDON_ID_PREFIX: &str = "yt_id:";
+pub const INTRO_FEED_ID: &str = "feed-intro";
```
elpiel: we should remove this.

src/constants.rs
```diff
@@ -47,6 +50,8 @@ lazy_static! {
     pub static ref STREAMING_SERVER_URL: Url =
         Url::parse("http://127.0.0.1:11470").expect("STREAMING_SERVER_URL parse failed");
     pub static ref IMDB_URL: Url = Url::parse("https://imdb.com").expect("IMDB_URL parse failed");
+    /// Meta hub url used to show background images for Notifications.
+    pub static ref METAHUB_URL: Url = Url::parse("https://images.metahub.space").expect("METAHUB_URL parse failed");
```
elpiel: This is not needed, since we shouldn't hardcode the url for addons but leave them to provide the poster.

PR #427, opened 2023-01-25
refactor(stream): remove tracker: and dht: from announce entries

tymmesyde:
Related https://github.com/Stremio/stremio-core/pull/421

### DIFF
--- a/src/types/resource/stream.rs
+++ b/src/types/resource/stream.rs
@@ -47,19 +47,19 @@ impl Stream {
                 xl: None,
                 tr: announce
                     .iter()
+                    // `tracker` and `dht` prefixes are used internally by the server.js
+                    // we need to remove those prefixes when generating the magnet URL
                     .map(|tracker| {
-                        if tracker.starts_with("tracker:") {
-                            tracker.replacen("tracker:", "", 1)
-                        } else {
-                            tracker.to_owned()
-                        }
+                        tracker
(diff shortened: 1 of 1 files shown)


src/types/resource/stream.rs
```diff
@@ -47,7 +47,20 @@ impl Stream {
                 xl: None,
                 tr: announce
                     .iter()
-                    .map(|tracker| tracker.replace("tracker:", ""))
+                    .map(|tracker| {
```
elpiel, line 52:
```suggestion
                    /// `tracker` and `dht` prefixes are used internally by the server.js
                    /// we need to remove those prefixes when generating the magnet URL
                    .map(|tracker| {
```

src/types/resource/stream.rs
```diff
@@ -47,7 +47,20 @@ impl Stream {
...
+                        if tracker.starts_with("tracker:") {
+                            tracker.replacen("tracker:", "", 1)
+                        } else {
+                            tracker.to_owned()
+                        }
```
elpiel: as a note, you can also use [strip_prefix](https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix)

E.g.:
`tracker.strip_prefix("tracker:").map(ToString::to_string).ok_or_else(tracker)`

elpiel (approved):

PR #430, opened 2023-02-16
handle player push to api behaviour internally

TheBeastLT:
Handle `push_to_api` behavior internally, instead of each individual implementation doing something differently and missing some push or having them too often.
Now it will push to api when updating time and check if at least 30s passed since the last update. As `update_time` is called on every video tick it should work just fine. 
Also push to api when stopping the video and when unloading the model.

### DIFF
--- a/src/models/player.rs
+++ b/src/models/player.rs
@@ -77,7 +77,7 @@ pub struct Player {
     #[serde(skip_serializing)]
     pub load_time: Option<DateTime<Utc>>,
     #[serde(skip_serializing)]
-    pub api_push_time: DateTime<Utc>,
+    pub push_library_item_time: DateTime<Utc>,
     #[serde(skip_serializing)]
     pub loaded: bool,
     #[serde(skip_serializing)]
@@ -225,7 +225,7 @@ impl<E: Env + 'static> UpdateWithCtx<E> for Player {
                 };
                 let switch_to_next_video_effects =
                     switch_to_next_video(&mut self.library_item, &self.next_video);
-                let push_to_api_effect = push_to_api(&self.library_item);
+                let push_to_library_effects = push_to_library(&self.library_item);
                 let selected_effects = eq_update(&mut self.selected, None);
                 let meta_item_effects = eq_update(&mut self.meta_item, None);
                 let subtitles_effects = eq_update(&mut self.subtitles, vec![]);
@@ -240,7 +240,7 @@ impl<E: Env + 'static> UpdateWithCtx<E> for Player {
                 self.ended = false;
                 self.paused = None;
                 switch_to_next_video_effects
-                    .join(push_to_api_effect)
+                    .join(push_to_library_effects)
                     .join(selected_effects)
                     .join(meta_item_effects)
                     .join(subtitles_effects)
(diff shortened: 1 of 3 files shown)


src/models/player.rs
```diff
@@ -334,15 +338,26 @@ impl<E: Env + 'static> UpdateWithCtx<E> for Player {
                         }
                     } else {
                         Effects::none()
-                    }
+                    };
+                    let push_to_api_effect = if self.api_push_time.is_none()
+                        || E::now() - self.api_push_time.unwrap() >= Duration::seconds(30)
```
elpiel: I'm not sure  `unwrap()` here is justified. What if the DateTime is None?
When do we actually set this DateTime and if we expect for it to always be there, then it should not be `Option`.

elpiel (changes requested):

src/models/player.rs
elpiel: Also, this should be `match`ed, not `if is_none || .unwrap()`

another way would be to `map()` the Option and and `unwrap_or_default(true)` so if no `api_push_time` was set before this, it would always go into this branch.

TheBeastLT: > When do we actually set this DateTime

Yes, forgot to update it.

> Also, this should be matched, not if is_none || .unwrap()

I've changed it to use `unwrap_or_default`, cuz that matches both conditions and is more concise than using `match`.
Or maybe it would be even better to declare `pub api_push_time: DateTime<Utc>,` and it would then use Default value of Epoch for it, then the condition could be just `E::now() - self.api_push_time >= Duration::seconds(30)`

src/models/player.rs
```diff
@@ -464,6 +474,16 @@ impl<E: Env + 'static> UpdateWithCtx<E> for Player {
     }
 }
 
+fn push_to_library(library_item: &Option<LibraryItem>) -> Effects {
```
elpiel: To follow convention it should be named: `update_library_item`

src/models/player.rs
```diff
@@ -77,6 +77,8 @@ pub struct Player {
     #[serde(skip_serializing)]
     pub load_time: Option<DateTime<Utc>>,
     #[serde(skip_serializing)]
+    pub push_library_item_time: DateTime<Utc>,
```
elpiel, line 92: I see where we replace the value with the new `DateTime` (`E::now()`) but when is this value set initially?

I'm still hesitant of whether or not we should use `Option`

TheBeastLT: It's set from the Default trait, which is Epoch. Each model is created via Model::new constructor. This is done in the core-web and core-kotlin.

elpiel: Right now DateTime does not support Default, meaning that this does not compile.
We should set the default value ourselves or use `Option`

TheBeastLT: DateTime does impl Default https://docs.rs/chrono/latest/chrono/struct.DateTime.html

elpiel: I've added derivative with default Timestamp of the Epoc, this it make sure we update the `LibraryItem` on `TimeChanged`

elpiel (approved):

PR #431, opened 2023-02-23
Basic impl of NotificationsBucket

### LINKED ISSUE #12: Stop showing items in continue watching section when fixed % of movie has been watched
**Description**
Movies should be removed from continue watching section when user has watched >= 95% of duration time

This task affects one model:

* LibRecent (continue watching)

--

Alternatively, for better behavior with series, we should implement this in the Player model: when the user stops playing something, if it's over a threshold, consider it played and try setting the `video_id` to the next one (if any) 

This way, when you watch 95% of an episode, and stop it, stremio will auto-set itself to the next one

This is currently implemented in https://github.com/Stremio/stremio/blob/development/src/controllers/player.js#L400

This involves the following models

* Player

### LINKED ISSUE #419: Notifications and continue watching merging
### Notifications

Notifications should be implemented in core.

Here's what Alex said about this:

> https://github.com/Stremio/stremio-api-workers/blob/master/notifItem/updateFeeds.js
> i think here is the better place
> https://github.com/Stremio/stremio-api-workers/blob/7bf55fa0b2eed7fcf0d04ca8d27efe1f258c0603/notifItem/updateFeeds.js#L31-L37
> here specifically, it is doing a last videos request to cinemeta
> like this: https://v3-cinemeta.strem.io/catalog/series/last-videos/lastVideosIds=tt7440726,tt0944947.json
> with multiple imdb id
> (i think 50 IDs is the limit for last-videos)
> it gets those IDs from the library items of the user
> it then analyses the response to figure out if there are any new videos for those series

## Continue watching

After implementing the Notifications based on the JS impl we should be able to merge the Notifications with the Continue watching items and create a single Continue watching bar with new episodes of series showing in front when they come out.

Related tasks:
- [ ] #12

unclekingpin:
still needs:
- schema migrations
- react to LibraryChanged and remove notifs
- move notifications_disabled from LibraryItemState to LibraryItem
- refactor VideoSortedVecAdapter to use meta_item.type instead of iterating the videos array to determine if its a series or not


elpiel:
>     * react to LibraryChanged and remove notifs
There are a few things we need to put more thought into:

- If you remove an item from your library but you still haven't watched the full movie/episode it should still stay in CW.
- If you remove an item from your library but you've also watched it already then it should be removed from CW.
- For the notifications themselves, they should be remove, I guess?

>     * schema migrations
>
>     * move notifications_disabled from LibraryItemState to LibraryItem

Is this really necessary?
Also, Is this the reason we need to make a new schema migration?

>     * refactor VideoSortedVecAdapter to use meta_item.type instead of iterating
> the videos array to determine if its a series or not

That's minor improvement we can look into.

src/types/library/library_item.rs
```diff
@@ -40,11 +46,22 @@ impl LibraryItem {
     pub fn should_sync<E: Env + 'static>(&self) -> bool {
         let year_ago = E::now() - Duration::days(365);
         let recently_removed = self.removed && self.mtime > year_ago;
-        self.r#type != "other" && (!self.removed || recently_removed)
+        self.r#type != MetaType::Other && (!self.removed || recently_removed)
     }
     #[inline]
     pub fn is_in_continue_watching(&self) -> bool {
-        self.r#type != "other" && (!self.removed || self.temp) && self.state.time_offset > 0
+        self.r#type != MetaType::Other && (!self.removed || self.temp) && self.state.time_offset > 0
```
elpiel, line 53: We should not include the LibraryItem in CW if we've watched a certain amount of the Movie/series.

@jaruba mentioned that this is working very well in the Android TV App so we should look into adding the same logic here as well.

unclekingpin:
Closed in favor of #433

unclekingpin:
- If you remove an item from your library but you still haven't watched the full movie/episode it should still stay in CW.
That could happen if we set temp to true in ActionCtx::RemoveFromLibrary when this condition is met. but seems like it should be addressed in seaprate PR.

unclekingpin:
- move notifications_disabled from LibraryItemState to LibraryItem
Ideally LibraryItemState should contain data about the watching progress of the item, which is modified by the player. We have to put all the other props about the libraryItem like removed, mtime etc in the root. The data migrations is a must because 2 props of the LibraryItemState ware renamed

unclekingpin:
* refactor VideoSortedVecAdapter to use meta_item.type instead of iterating
if you can do that it will be great. i tried and couldnt do it. Perhaps the serde_with extention must be applied to MetaItem, not just the videos, so that it has access to the type?

PR #433, opened 2023-03-12
update NotificationsBucket impl

unclekingpin:


src/models/ctx/update_notifications.rs
```diff
@@ -0,0 +1,213 @@
...
+                {
+                    meta_item
+                        .videos_iter()
+                        .find(|video| {
+                            match (&library_item.state.last_video_released, &video.released) {
+                                (Some(last_video_released), Some(video_released)) => {
+                                    video_released > last_video_released
+                                }
+                                _ => false,
+                            }
```
elpiel, line 184: - [ ] You need to exclude future videos

src/models/ctx/update_notifications.rs
```diff
@@ -0,0 +1,213 @@
...
+                        })
+                        .map(|video| NotificationItem {
+                            id: id.to_owned(),
+                            video: video.to_owned(),
+                        })
```
elpiel, line 189: Having a single video for a given LIbraryItem could be a problem because we might end up showing a number with the new episodes.

src/types/resource/meta_item.rs
```diff
@@ -198,6 +199,20 @@ pub struct MetaItem {
...
+        if self.preview.r#type == "series" {
+            Either::Left(self.videos.iter().filter(|video| match video.series_info {
+                Some(SeriesInfo { season, .. }) => season != 0,
+                _ => false,
+            }))
+        } else {
+            Either::Right(self.videos.iter().rev())
+        }
+    }
+}
```
elpiel, line 214: - [ ] Please add the 2 cases - when `series` and when not `series` in the documentation

PR #435, opened 2023-04-06
Local search for search bar suggestions

### LINKED ISSUE #434: Add Local search
Implement the local search functionality in core using the https://github.com/Stremio/local-search/ crate developed for this application.

The local search functionality is used for the search bar only to suggest MetaItems when user starts typing before the actual search is performed.

**Requires** a suitable feed from cinemeta: https://github.com/Stremio/cinemeta-catalogs/issues/14

Related: https://github.com/Stremio/stremio-core/issues/148
PRs:
- https://github.com/Stremio/stremio-core/pull/435

elpiel:
Closes #434 

Testing on web (console):

```js
await services.core.transport.dispatch({
    action: 'Search',
      args: {
        action: 'Search',
        args: {
          searchQuery: "Ant man",
          maxResults: 5
        }
    }
})
```

you can then check the results by:

```js
await services.core.transport.getState("local_search")
```

and checking out the `search_results` field


src/models/local_search.rs
```diff
@@ -0,0 +1,405 @@
+//! Local autocompletion search
```
TheBeastLT, line 1: Is this needed?

src/models/local_search.rs
```diff
@@ -0,0 +1,405 @@
...
+#[serde(transparent)]
+pub struct SearchableItemsResponse(pub Vec<Searchable>);
+
+#[derive(Copy, Clone)]
+pub struct IndexOptions {
+    imdb_rating_weight: f64,
+    popularity_weight: f64,
+}
+
+/// A searchable item
```
TheBeastLT, line 46: Do such comments bring any value?

src/models/local_search.rs
```diff
@@ -0,0 +1,405 @@
...
+    pub release_info: Option<String>,
+}
+
+/// Local search functionality for the search engine's suggestions when typing
+#[derive(Default, Serialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct LocalSearch {
+    /// The Searchable items that will be used for the local search.
+    ///
+    // #[serde(skip)]
```
TheBeastLT: Commented code?

TheBeastLT (approved):
LGTM. Tho not sure about the value of some comments, some seem just redundant.

src/models/local_search.rs
elpiel: It's good for documentation when you're looking at e.g. https://stremio.github.io/stremio-core/stremio_core/models/index.html

src/models/local_search.rs
elpiel: Good catch, this should be uncommented, it was only for testing.

src/models/local_search.rs
elpiel: yes, because we need to define what this struct does and document all fields.

src/runtime/msg/action.rs
```diff
@@ -119,12 +119,24 @@ pub enum ActionLoad {
...
+    LocalSearch,
     MetaDetails(MetaDetailsSelected),
     Player(Box<PlayerSelected>),
     Link,
     Notifications,
 }
 
+#[derive(Clone, Deserialize, Debug)]
+#[serde(tag = "action", content = "args")]
+pub enum ActionSearch {
```
elpiel, line 132: - [x] Should we rename all fields to `camelCase`?

PR #438, opened 2023-04-25
replace frame rate matching bool flag with an enum

TheBeastLT:
Depending on a device we might want to have different strategies when matching display refresh rate to video frame rate.
On most TVs and boxes it makes sense to match the frame rate and the resolution of the video, so that if the TV could upscale the content if it has the functionality for it.
In other cases (atm I think it's just Nvidia shield), some TV boxes do the upscaling on the device and it needs to retain the resolution to which it should upscale (usually 4k).
So that's why a bool flag is no longer a good option and I think an enum should fir this purpose better.


src/runtime/env.rs
```diff
@@ -414,10 +420,36 @@ fn migrate_storage_schema_to_v6<E: Env>() -> TryEnvFuture<()> {
...
+    E::get_storage::<serde_json::Value>(PROFILE_STORAGE_KEY)
+        .and_then(|mut profile| {
+            match profile
+                .as_mut()
+                .and_then(|profile| profile.as_object_mut())
+                .and_then(|profile| profile.get_mut("settings"))
+                .and_then(|settings| settings.as_object_mut())
+            {
+                Some(settings) => {
+                    settings.remove("autoFrameRateMatching");
```
TheBeastLT, line 433: Here I'm just removing the value, as it was not used in any of the apps.

src/runtime/env.rs
```diff
@@ -463,7 +523,7 @@ mod test {
             set_profile_and_schema_version(&profile_before, 5);
 
             // migrate storage
-            TestEnv::migrate_storage_schema()
+            migrate_storage_schema_to_v6::<TestEnv>()
```
TheBeastLT, line 526: Changed to test the specific migration method, as otherwise you'd need to change each test after adding a new migration, since it applies all of them down the list.

src/runtime/env.rs
```diff
@@ -440,6 +472,34 @@ mod test {
...
+
+            // migrate storage
+            TestEnv::migrate_storage_schema()
+                .await
+                .expect("Should migrate");
+
+            let storage = STORAGE.read().expect("Should lock");
+
+            assert_eq!(
+                &SCHEMA_VERSION.to_string(),
```
elpiel, line 494: I believe this should now be 7?
I would suggest to do as the other tests - name the test `v6 to v7` and use hardcoded value instead of `SCHEMA_VERSION`. This will eliminate the need for this test to be update on a new schema.

elpiel (approved):
LGTM! You can freely merge when you address my comment regarding the tests.

src/runtime/env.rs
TheBeastLT: This test is specifically like this so you wouldn't need to change it and it would test that the migrations really update to the latest defined version, as this runs the whole migration process not a single migration method.

elpiel: Ok, got you! This will require a change in the test once we have a new migration too though. Something to keep in mind and look to improve in the future.

TheBeastLT: No, you shouldn't need to change anything in this test when adding a new version if you do everything correctly.
This will test and fail if you:
 - create a new migration and bump the schema version in the migration, but forget to update the const `SCHEMA_VERSION`
 - you bump the `SCHEMA_VERSION` but forget to create a migration for the new schema version
 - you create the migration and bump the `SCHEMA_VERSION`, but in your migration you've forgot to update the schema version.

PR #441, opened 2023-04-28
improve addon upgrade behaviour

TheBeastLT:
Currently when upgrading addon it checks for installed addons based on `id` and replaces all of them with the new one. But this is flawed, since addons might have the same `id`, but different configuration and that's not what the user would expect, he'd expect to upgrade the same addon with the same configuration to the new manifest version. So I'm updating this function to reflect that.
Also fix a bug which allowed upgrading `protected` addon, which would add a second addon with the same `transportUrl`. Same was possible for uninstall, as I had a case where in my profile `cinemeta` was stored without `protected` variable and was able to uninstall it, so check both the local and passed addon flags as usually the passed addon would be the remote one, which would have the flag.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -72,6 +72,7 @@ lazy_static! {
         options_limit: OptionsLimit::default(),
     };
     pub static ref TYPE_PRIORITIES: HashMap<&'static str, i32> = vec![
+        ("all", 5),
         ("movie", 4),
         ("series", 3),
         ("channel", 2),
--- a/src/models/ctx/update_profile.rs
+++ b/src/models/ctx/update_profile.rs
@@ -107,18 +107,19 @@ pub fn update_profile<E: Env + 'static>(
             if addon.manifest.behavior_hints.configuration_required {
                 return addon_upgrade_error_effects(addon, OtherError::AddonConfigurationRequired);
             }
-            let addon_position = profile
+            let addon_position = match profile
                 .addons
                 .iter()
                 .map(|addon| &addon.transport_url)
-                .position(|transport_url| *transport_url == addon.transport_url);
-            if addon_position.is_none() {
-                return addon_upgrade_error_effects(addon, OtherError::AddonNotInstalled);
-            }
-            if addon.flags.protected || profile.addons[addon_position.unwrap()].flags.protected {
+                .position(|transport_url| *transport_url == addon.transport_url)
+            {
+                Some(addon_position) => addon_position,
+                None => return addon_upgrade_error_effects(addon, OtherError::AddonNotInstalled),
+            };
+            if addon.flags.protected || profile.addons[addon_position].flags.protected {
                 return addon_upgrade_error_effects(addon, OtherError::AddonIsProtected);
             }
-            profile.addons[addon_position.unwrap()] = addon.to_owned();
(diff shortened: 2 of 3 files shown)


src/unit_tests/ctx/upgrade_addon.rs
```diff
@@ -32,12 +32,31 @@ fn actionctx_installaddon_upgrade() {
             addon_catalogs: vec![],
             behavior_hints: Default::default(),
         },
-        transport_url: Url::parse("https://transport_url1").unwrap(),
+        transport_url: Url::parse("https://transport_url").unwrap(),
         flags: Default::default(),
     };
     let addon2 = Descriptor {
```
elpiel: we can clarify that this is update of addon1, e.g. `addon1_update`

src/models/ctx/update_profile.rs
```diff
@@ -101,54 +101,37 @@ pub fn update_profile<E: Env + 'static>(
...
+            }
+            let addon_position = profile
+                .addons
+                .iter()
+                .map(|addon| &addon.transport_url)
+                .position(|transport_url| *transport_url == addon.transport_url);
+            if addon_position.is_none() {
+                return addon_upgrade_error_effects(addon, OtherError::AddonNotInstalled);
+            }
+            if addon.flags.protected || profile.addons[addon_position.unwrap()].flags.protected {
```
elpiel: I would avoid `unwrap` even though it safe here as we could do it better with a match on the `addon_position` and remove the need for `unwrap()` all together.

elpiel:
```suggestion
            let mut profile_addon = match profile
                .addons
                .iter_mut()
                .find(|profile_addon| *transport_url == profile_addon.transport_url) {
                    Some(profile_addon) => profile_addon,
                    None => return addon_upgrade_error_effects(addon, OtherError::AddonNotInstalled),
                };

            if addon.flags.protected || profile_addon.flags.protected {
```

it might not be possible to replace the full addon in the profile when using `iter_mut`   and `*profile_addon = addon` but you can try anyway.

src/constants.rs
```diff
@@ -72,6 +72,7 @@ lazy_static! {
         options_limit: OptionsLimit::default(),
     };
     pub static ref TYPE_PRIORITIES: HashMap<&'static str, i32> = vec![
+        ("all", 5),
```
elpiel, line 75: Is this necessary for the PR? I don't see it used anywhere.

elpiel (commented):
LGMT, a few remarks regarding tests & simplifying code and is good to go.

Also, for next time, please make sure to create a PR towards `development` to fix things like clippy warning. This will make sure that we don't get locked to a PR like this one to fix the clippy warnings while all other PRs are experiencing the same warnings.

src/constants.rs
TheBeastLT: Yes, this is needed when listing addon catalogs, since this type should be the first in the list, but actually is sorted after all the priority types.
image
It a small change so thought I'd include it here with a related the addon action as well.

src/models/ctx/update_profile.rs
TheBeastLT: I wasn't able to replace the full object, but updated to use matching to find the position instead.

elpiel (approved):

PR #445, opened 2023-05-03
increase continue watching returned item count

TheBeastLT:
Since we don't load all catalogs at the same time and are they loaded as we scroll the board, we can increase the CW items count displayed to the same number as addon catalog page.


elpiel (approved):
LGTM! This should not affect the performance of the apps, however, each app should make sure to show as many items as the app requires.

PR #446, opened 2023-05-03
Addons pulled from API args

unclekingpin:
send diff of transport_urls in AddonsPulledFromAPI event args


src/models/ctx/update_profile.rs
```diff
@@ -82,10 +83,14 @@ pub fn update_profile<E: Env + 'static>(
...
-                    .cloned()
+                let local_addons = profile.addons.iter().cloned().collect::<HashSet<_>>();
+                let remote_addons = next_addons.iter().cloned().collect::<HashSet<_>>();
+                let added_addons = &remote_addons - &local_addons;
+                let removed_addon = &local_addons - &remote_addons;
+                let transport_urls = added_addons
+                    .into_iter()
+                    .chain(removed_addon.into_iter())
+                    .map(|addon| addon.transport_url)
                     .collect();
```
elpiel: Should we try to deduplicate the result to remove duplicates?

elpiel (approved):
I only have 1 small comment for the deduplication and I would like to discuss it before merging.

src/models/ctx/update_profile.rs
```diff
@@ -82,10 +83,22 @@ pub fn update_profile<E: Env + 'static>(
...
                     .iter()
                     .map(|addon| &addon.transport_url)
                     .cloned()
+                    .collect::<HashSet<_>>();
+                let added_transport_urls = &next_transport_urls - &prev_transport_urls;
+                let removed_transport_urls = &prev_transport_urls - &next_transport_urls;
+                let transport_urls = added_transport_urls
+                    .into_iter()
+                    .chain(removed_transport_urls.into_iter())
                     .collect();
```
elpiel, line 102: This could be moved to it's own function as it repeats with the lines below.

elpiel (approved):

src/models/ctx/update_profile.rs
```diff
@@ -82,10 +83,22 @@ pub fn update_profile<E: Env + 'static>(
...
+                    .map(|addon| &addon.transport_url)
+                    .cloned()
+                    .collect::<HashSet<_>>();
+                let next_transport_urls = next_addons
                     .iter()
                     .map(|addon| &addon.transport_url)
                     .cloned()
+                    .collect::<HashSet<_>>();
+                let added_transport_urls = &next_transport_urls - &prev_transport_urls;
+                let removed_transport_urls = &prev_transport_urls - &next_transport_urls;
```
elpiel, line 98: Why do we need to add the removed `transport_url`s to the `AddonsPulledFromAPI` event?

unclekingpin: the idea is to receive the **changed** transport urls in the event arg. so we know that if the vec is not empty something was changed

PR #450, opened 2023-05-12
Notifications bucket part 2

elpiel:
Continuation of the work that @unclekingpin has put up in #431


src/models/player.rs
```diff
@@ -223,6 +223,13 @@ impl<E: Env + 'static> UpdateWithCtx<E> for Player {
                 );
                 let watched_effects =
                     watched_update(&mut self.watched, &self.meta_item, &self.library_item);
+                let notification_effects = match &selected.meta_request {
+                    Some(meta_request) => Effects::msg(Msg::Internal(
+                        Internal::DismissNotificationItem(meta_request.path.id.to_owned()),
```
elpiel, line 228: Plane wrong.
Right now the notification bucket stores the Videos and uses the video ID as key while this like sends a MetaItem id.

elpiel: latest code uses MetaItemId for the key of the first HashMap. This is not correct.

PR #452, opened 2023-05-15
fix video params logic

TheBeastLT:
When `video_params` were added, the logic assumed that it will always be provided and when it was not provided it wouldn't query any of the subtitles addons. `video_params` won't always be available (stremio-web without streaming server for example), so it should not prevent querying subtitle addons with just the `videoId`.
Also these new extra values were not propagated to the legacy transport subtitle method.

### DIFF
--- a/src/addon_transport/http_transport/legacy/mod.rs
+++ b/src/addon_transport/http_transport/legacy/mod.rs
@@ -210,13 +210,13 @@ fn build_legacy_req(transport_url: &Url, path: &ResourcePath) -> Result<Request<
                     serde_json::Value::String(video_hash.to_owned()),
                 );
             }
-            let video_size = path.get_extra_first_value(VIDEO_SIZE_EXTRA_PROP.name.as_str());
+            let video_size = path
+                .get_extra_first_value(VIDEO_SIZE_EXTRA_PROP.name.as_str())
+                .and_then(|video_size| video_size.parse().ok());
             if let Some(video_size) = video_size {
                 query.insert(
                     VIDEO_SIZE_EXTRA_PROP.name.as_str(),
-                    serde_json::Value::Number(
-                        video_size.parse().expect("Failed to parse videoSize"),
-                    ),
+                    serde_json::Value::Number(video_size),
                 );
             }
             build_jsonrpc("subtitles.find", json!({ "query": query }))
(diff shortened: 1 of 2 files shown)


src/addon_transport/http_transport/legacy/mod.rs
```diff
@@ -199,10 +200,27 @@ fn build_legacy_req(transport_url: &Url, path: &ResourcePath) -> Result<Request<
...
+                    VIDEO_HASH_EXTRA_PROP.name.as_str(),
+                    serde_json::Value::String(video_hash.to_owned()),
+                );
+            }
+            let video_size = path.get_extra_first_value(VIDEO_SIZE_EXTRA_PROP.name.as_str());
+            if let Some(video_size) = video_size {
+                query.insert(
+                    VIDEO_SIZE_EXTRA_PROP.name.as_str(),
+                    serde_json::Value::Number(
+                        video_size.parse().expect("Failed to parse videoSize"),
```
elpiel: I would suggest not to `expect` this value after parsing.
I would instead suggest to skip the `video_size` if it's invalid (or `unwrap_default()`)

elpiel (changes requested):
LGTM, I only have this one tweak for the `.expect` that should be fixed.
In general, do we want to send a default value or not send the property at all if it's invalid?
In this case I'm talking about `video size`.

src/addon_transport/http_transport/legacy/mod.rs
TheBeastLT: I'd say that it's better to fail in this case as this is not expected, since under normal circumstances it can't happen, as the `VideoParams` struct has videoSize as `uint`. And if we just silently swallow and not send it, we might have a bug that we will ignore and not notice for a long time.

elpiel: And if you have that bug - app crashes and you cannot use it.

elpiel: We should log such errors but we must not crash the app if there's a bug or bad input

TheBeastLT: I'd still think `expect` is better here, because this is a controlled flow only done within the `stremio-core` codebase as the input will always be a number, but fair enough.

PR #461, opened 2023-06-03
feat: library item progress

tymmesyde:
Add `stream_id` to LibraryItemState 
Related https://github.com/Stremio/stremio-web/issues/383

- `core-web` changes: https://github.com/Stremio/stremio-core-web/pull/51


src/types/library/library_item.rs
```diff
@@ -112,6 +120,10 @@ pub struct LibraryItemState {
     #[serde(default, rename = "video_id")]
     #[serde_as(deserialize_as = "DefaultOnNull<NoneAsEmptyString>")]
     pub video_id: Option<String>,
+    /// The last stream watched.
+    #[serde(default, rename = "stream_id")]
+    #[serde_as(deserialize_as = "DefaultOnNull<NoneAsEmptyString>")]
+    pub stream_id: Option<String>,
```
elpiel: Let's keep the camelCase. This is very weird, by default we camelCase the fields, but on this and the `video_id` we use snake_case?!
This should be fixed, we need consistency.

elpiel (changes requested):

src/types/library/library_item.rs
tymmesyde: I kept it because the other fields are snakecase, it's related to the way the api works or was implemented idk

src/models/ctx/update_library.rs
```diff
@@ -288,7 +288,16 @@ fn push_items_to_api<E: Env + 'static>(items: Vec<LibraryItem>, auth_key: &AuthK
         fetch_api::<E, _, _, SuccessResponse>(&DatastoreRequest {
             auth_key: auth_key.to_owned(),
             collection: LIBRARY_COLLECTION_NAME.to_owned(),
-            command: DatastoreCommand::Put { changes: items },
+            command: DatastoreCommand::Put {
+                changes: items
+                    .iter()
```
elpiel: if you use `into_iter` instead (because either way you have an owned `items: Vec<..>` you can also avoid the `clone()` from line 295 (you should have `|mut item|` for the closure

src/models/ctx/update_library.rs
```diff
@@ -288,7 +288,16 @@ fn push_items_to_api<E: Env + 'static>(items: Vec<LibraryItem>, auth_key: &AuthK
...
+                    .map(|item| {
+                        let mut item_without_stream = item.clone();
+                        item_without_stream.state.stream = None;
+                        item_without_stream
```
elpiel: - [ ] If we end up doing this  instead of skipping the field in serialization, please document very well in a comment why we remove the `stream` - we do not want to save the stream that the user is playing in our DB.

src/types/library/library_item.rs
```diff
@@ -46,6 +46,14 @@ impl LibraryItem {
     pub fn is_in_continue_watching(&self) -> bool {
         self.r#type != "other" && (!self.removed || self.temp) && self.state.time_offset > 0
     }
+    #[inline]
+    pub fn progress(&self) -> f64 {
+        if self.state.time_offset > 0 && self.state.duration > 0 {
+            self.state.time_offset as f64 / self.state.duration as f64
```
elpiel: Depending on what values we expect here, for large-enough values (like `u64::MAX`) we might get into problems when we use `as f64`:

https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=6e6409d3fd006ac95e1c907299bc826a

Do we have to worry about such large values?
If we do need, then I would suggest using `Ratio`:

https://docs.rs/num/latest/num/rational/struct.Ratio.html#impl-ToPrimitive
And then try to make a f64 out of it:

https://docs.rs/num/latest/num/trait.ToPrimitive.html#method.to_f64

src/types/library/library_item.rs
```diff
@@ -46,6 +46,14 @@ impl LibraryItem {
     pub fn is_in_continue_watching(&self) -> bool {
         self.r#type != "other" && (!self.removed || self.temp) && self.state.time_offset > 0
     }
+    #[inline]
+    pub fn progress(&self) -> f64 {
```
elpiel, line 58: - [x] Replace this progress calculation in `core-web`:
https://github.com/Stremio/stremio-core-web/blob/development/src/model/serialize_library.rs#L110-L114

Done in: https://github.com/Stremio/stremio-core-web/pull/51

src/types/library/library_item.rs
```diff
@@ -112,6 +120,10 @@ pub struct LibraryItemState {
     #[serde(default, rename = "video_id")]
     #[serde_as(deserialize_as = "DefaultOnNull<NoneAsEmptyString>")]
     pub video_id: Option<String>,
+    /// The last stream watched.
+    #[serde(default)]
+    #[serde_as(deserialize_as = "DefaultOnNull<NoneAsEmptyString>")]
+    pub stream: Option<String>,
```
elpiel: Instead of setting the `stream` to `None` on DataStore::Put, we can instead `#[serde(skip_serialization)]` for the field.
- [ ] TODO: Check if this serialization scheme is intended for all use cases of the LibraryItemStatus

- https://github.com/Stremio/stremio-core-web/blob/af66b2eb6bd1aae35e18cdf66e70879100ee7fd5/src/model/serialize_player.rs#L67-L73
- https://github.com/Stremio/stremio-core-web/blob/af66b2eb6bd1aae35e18cdf66e70879100ee7fd5/src/model/serialize_continue_watching_preview.rs#L11-L15

unclekingpin:
@tymmesyde Should we use the new stream bucket now?
cc: @elpiel

elpiel:
> @tymmesyde Should we use the new stream bucket now? cc: @elpiel

Yes, we should.

elpiel (approved):

elpiel (approved):

PR #466, opened 2023-06-06
Streams bucket patch

unclekingpin:
Based on #337 
- fixed conflicts
- refactored to not handle player events in ctx
- Should we allow multiple streams per mata_id/video_id?
- Should we implement a cleanup function in this pr?
- Should we add duration/offset or progress ratio to the StreamItem?

### DIFF
--- a/src/models/ctx/update_streams.rs
+++ b/src/models/ctx/update_streams.rs
@@ -1,0 +1,75 @@
+use enclose::enclose;
+use futures::FutureExt;
+
+use crate::constants::STREAMS_STORAGE_KEY;
+use crate::models::ctx::{CtxError, CtxStatus};
+use crate::runtime::msg::{Action, ActionCtx, Event, Internal, Msg};
+use crate::runtime::{Effect, EffectFuture, Effects, Env, EnvFutureExt};
+use crate::types::streams::{StreamsBucket, StreamsItem};
+
+pub fn update_streams<E: Env + 'static>(
+    streams: &mut StreamsBucket,
+    status: &CtxStatus,
+    msg: &Msg,
+) -> Effects {
+    match msg {
+        Msg::Action(Action::Ctx(ActionCtx::Logout)) | Msg::Internal(Internal::Logout) => {
+            let next_streams = StreamsBucket::default();
+            if *streams != next_streams {
+                *streams = next_streams;
+                Effects::msg(Msg::Internal(Internal::StreamsChanged(false)))
+            } else {
+                Effects::none().unchanged()
+            }
+        }
+        Msg::Internal(Internal::StreamLoaded(
+            Some(meta_item_id),
+            Some(video_id),
+            Some(transport_url),
+            stream,
+        )) => {
+            let streams_item = StreamsItem {
+                stream: stream.to_owned(),
+                transport_url: transport_url.to_owned(),
+                mtime: E::now(),
+            };
+            streams
+                .items
+                .insert((meta_item_id.to_owned(), video_id.to_owned()), streams_item);
+            Effects::msg(Msg::Internal(Internal::StreamsChanged(false)))
+        }
+        Msg::Internal(Internal::StreamsChanged(persisted)) if !persisted => {
+            Effects::one(push_streams_to_storage::<E>(streams)).unchanged()
+        }
+        Msg::Internal(Internal::CtxAuthResult(auth_request, result)) => match (status, result) {
+            (CtxStatus::Loading(loading_auth_request), Ok((auth, _, _)))
+                if loading_auth_request == auth_request =>
+            {
+                let next_streams = StreamsBucket::new(Some(auth.user.id.to_owned()));
+                if *streams != next_streams {
+                    *streams = next_streams;
+                    Effects::msg(Msg::Internal(Internal::StreamsChanged(false)))
+                } else {
+                    Effects::none().unchanged()
+                }
+            }
+            _ => Effects::none().unchanged(),
+        },
+        _ => Effects::none().unchanged(),
+    }
+}
+
+fn push_streams_to_storage<E: Env + 'static>(streams: &StreamsBucket) -> Effect {
+    EffectFuture::Sequential(
+        E::set_storage(STREAMS_STORAGE_KEY, Some(&streams))
+            .map(enclose!((streams.uid => uid) move |result| match result {
+                Ok(_) => Msg::Event(Event::StreamsPushedToStorage { uid }),
+                Err(error) => Msg::Event(Event::Error {
+                    error: CtxError::from(error),
+                    source: Box::new(Event::StreamsPushedToStorage { uid }),
+                })
+            }))
+            .boxed_env(),
+    )
+    .into()
+}
--- a/src/models/ctx/update_streams.rs
+++ b/src/models/ctx/update_streams.rs
@@ -2,7 +2,7 @@ use crate::constants::STREAMS_STORAGE_KEY;
 use crate::models::ctx::{CtxError, CtxStatus};
 use crate::runtime::msg::{Action, ActionCtx, Event, Internal, Msg};
 use crate::runtime::{Effect, EffectFuture, Effects, Env, EnvFutureExt};
-use crate::types::streams::{StreamsBucket, StreamsItem};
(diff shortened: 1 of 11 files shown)


src/types/streams/streams_bucket.rs
```diff
@@ -0,0 +1,25 @@
...
+use crate::types::profile::UID;
+use crate::types::streams::StreamsItem;
+
+#[serde_as]
+#[derive(Default, Clone, PartialEq, Serialize, Deserialize)]
+#[cfg_attr(debug_assertions, derive(Debug))]
+pub struct StreamsBucket {
+    pub uid: UID,
+    #[serde_as(as = "Vec<(_, _)>")]
+    pub items: HashMap<(String, String), StreamsItem>,
```
elpiel: So this would be serialized as `Vec<((String, String), StreamItem)>`?

- [x] Please document what the keys of "items"  are

src/types/streams/streams_bucket.rs
```diff
@@ -0,0 +1,25 @@
...
+
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+
+use crate::types::profile::UID;
+use crate::types::streams::StreamsItem;
+
+#[serde_as]
+#[derive(Default, Clone, PartialEq, Serialize, Deserialize)]
+#[cfg_attr(debug_assertions, derive(Debug))]
```
elpiel: Small reminder that I did a comparison with these `derive(Debug)`s and we agreed that the size difference for wasm wasn't not worth it.
- [x] `derive(Debug)` and remove the `cfg_attr`

src/runtime/msg/internal.rs
```diff
@@ -43,12 +44,16 @@ pub enum Internal {
     Logout,
     /// Dispatched when addons needs to be installed.
     InstallAddon(Descriptor),
+    /// Dispatched when a new stream is loaded into the Player.
+    StreamLoaded(Option<String>, Option<String>, Option<Url>, Stream),
```
elpiel: Can we use Enums with fields for this, please?
Again for documentation and clarity we should strive to document and make readable the fields and attributes, so I suggest we use fields here and document each field.

src/types/streams/streams_bucket.rs
elpiel: Key is `(meta_item_id, video_id)`

src/types/streams/streams_item.rs
```diff
@@ -0,0 +1,15 @@
+use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+use url::Url;
+
+use crate::types::resource::Stream;
+
+#[serde_as]
+#[derive(Clone, PartialEq, Serialize, Deserialize)]
+#[cfg_attr(debug_assertions, derive(Debug))]
```
elpiel: - [x] Same as above - we should remove the `cfg_attr`

elpiel (commented):
Apart from the repeatability of some logic the PR and logic looks good.
Small improvements for documentation and derives are needed but the rest is Ok.


We have a lot of repeatability in the following form, we can improve this pattern for Effects:
```rust
            let next_streams = StreamsBucket::default();
            if *streams != next_streams {
                *streams = next_streams;
                Effects::msg(Msg::Internal(Internal::StreamsChanged(false)))
            } else {
                Effects::none().unchanged()
            }
```

src/types/streams/streams_item.rs
unclekingpin: seems like ive copy pasted from the wrong place. will fix

src/models/ctx/update_streams.rs
```diff
@@ -0,0 +1,78 @@
...
+            } else {
+                Effects::none().unchanged()
+            }
+        }
+        Msg::Internal(Internal::StreamLoaded {
+            stream,
+            meta_id: Some(meta_id),
+            video_id: Some(video_id),
+            transport_url: Some(transport_url),
+        }) => {
```
elpiel, line 29: For sanity check: If any of these fields is `None` we won't update the streams in the bucket.
This is the intended behaviour correct?

src/runtime/msg/internal.rs
```diff
@@ -43,12 +44,21 @@ pub enum Internal {
     Logout,
     /// Dispatched when addons needs to be installed.
     InstallAddon(Descriptor),
+    /// Dispatched when a new stream is loaded into the Player.
+    StreamLoaded {
+        stream: Stream,
+        meta_id: Option<String>,
+        video_id: Option<String>,
+        transport_url: Option<Url>,
+    },
```
elpiel, line 53: What are the use-cases for when we have `StreamLoaded` but no meta_id or video_id or transport_url?

src/types/streams/streams_item.rs
```diff
@@ -0,0 +1,15 @@
+use crate::types::resource::Stream;
+use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+use url::Url;
+
+#[serde_as]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct StreamsItem {
```
elpiel, line 10: We are missing:
```suggestion
#[serde(rename_all = "camelCase")]
pub struct StreamsItem {
```

src/types/streams/streams_item.rs
```diff
@@ -0,0 +1,15 @@
...
+    pub stream: Stream,
+    pub meta_id: String,
+    pub video_id: String,
+    pub transport_url: Url,
+    pub mtime: DateTime<Utc>,
```
elpiel, line 17: - [ ] Please add docs that this is the modify time and what it actually mean/when is updated.

src/types/streams/streams_bucket.rs
```diff
@@ -0,0 +1,28 @@
...
+pub struct StreamsItemKey {
+    pub meta_id: String,
+    pub video_id: String,
+}
+
+#[serde_as]
+#[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct StreamsBucket {
+    pub uid: UID,
+    #[serde_as(as = "Vec<(_, _)>")]
```
elpiel, line 18: What will be the json representation in this case?

I guess it will be in this form:
```json
[
    [
       { "meta_id": "...", "video_id": "..."},
       { ... StreamsItem ... }
    ]
]
```

So we need the fields in `camelCase` and wanted to check if this is the correct representation that we want for the `StreamsBucket.items`

elpiel (changes requested):
I think we're almost ready with this PR but it needs a few tweaks and documentation on a couple of places + I've added sanity check questions to validate the logic with you @unclekingpin

src/types/streams/streams_bucket.rs
unclekingpin: fixed

src/types/streams/streams_bucket.rs
unclekingpin: fixed

src/runtime/msg/internal.rs
unclekingpin: fixed

src/types/streams/streams_item.rs
unclekingpin: fixed

src/models/ctx/update_streams.rs
unclekingpin: correct! that happens when you want to watch a trailer. you dont want it to be part of the streams bucket

src/runtime/msg/internal.rs
unclekingpin: trailers
we can create /player deep link only with a stream and no metaid, videoid etc
https://github.com/unclekingpin/stremio-core/blob/development/src/deep_links/mod.rs#L296-L312

src/types/streams/streams_bucket.rs
unclekingpin: renamed the StreamsItemKey keys to camelCase
i think that is the correct representation, but it really doesnt matter as long as rust is being able to deserialize from it.
We will never process that serialized data with anything else so unique keys should be guaranteed

elpiel (approved):

PR #468, opened 2023-06-12
override meta details selected based on loaded meta

unclekingpin:
Override meta details selected based on viable meta if
1. meta_item has single video, use video id
2. meta_item has 0 videos and default video id, use default video id
3. meta_item has 0 videos and is a movie, use meta item id

Viable meta item is the first item in MetaDetails.meta_items that is not in Err state.

This pr does not require changes in core-web or web, not so sure about the tv app.

- [x] write unit tests
- [x] derive Clone for all models #474

### DIFF
--- a/src/models/meta_details.rs
+++ b/src/models/meta_details.rs
@@ -112,6 +112,13 @@ impl<E: Env + 'static> UpdateWithCtx<E> for MetaDetails {
                     &mut self.meta_items,
                     ResourcesAction::ResourceRequestResult { request, result },
                 );
+                let selected_effects: Effects =
+                    selected_override_update(&mut self.selected, &self.meta_items);
+                let streams_effects = if selected_effects.has_changed {
+                    streams_update::<E>(&mut self.streams, &self.selected, &ctx.profile)
+                } else {
+                    Effects::default()
+                };
                 let meta_streams_effects =
                     meta_streams_update(&mut self.meta_streams, &self.selected, &self.meta_items);
                 let library_item_effects = library_item_update::<E>(
@@ -122,8 +129,10 @@ impl<E: Env + 'static> UpdateWithCtx<E> for MetaDetails {
                 );
                 let watched_effects =
                     watched_update(&mut self.watched, &self.meta_items, &self.library_item);
-                meta_items_effects
+                selected_effects
+                    .join(meta_items_effects)
                     .join(meta_streams_effects)
+                    .join(streams_effects)
                     .join(library_item_effects)
                     .join(watched_effects)
             }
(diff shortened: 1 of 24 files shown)


src/models/meta_details.rs
```diff
@@ -189,6 +201,55 @@ fn library_item_sync(library_item: &Option<LibraryItem>, profile: &Profile) -> E
...
+) -> Effects {
+    let meta_path = if let Some(Selected {
+        meta_path,
+        stream_path: None,
+    }) = &selected
+    {
+        meta_path
+    } else {
+        return Effects::default();
+    };
```
elpiel: Nitpick: This is why in Rust we have a `match`, it doesn't make sense to use `if let` here.
```suggestion
    let meta_path = match &selected {
        Some(Selected { meta_path, .. }) => meta_path,
        None => return Effects::default(),
    }
```

TheBeastLT:
On Android TV we reload the details if `videoId` is missing when there's a `defaultVideoId` defined or the meta doesn't have any videos - https://github.com/Stremio/stremio-android-tv/blob/591861e946ad4f720fbc9a89cc4a3face07a946f/androidTV/src/main/kotlin/com/stremio/tv/viewmodels/MetaDetailsViewModel.kt#L154-L159
When there's only one video I don't think we should go to that video's details, as it might be a valid meta with just one video, and it's not done like this on desktop or other apps.

src/models/meta_details.rs
```diff
@@ -189,6 +201,55 @@ fn library_item_sync(library_item: &Option<LibraryItem>, profile: &Profile) -> E
...
+    let meta_item = if let Some(meta_item) = meta_items
+        .iter()
+        .find_map(|meta_item| match &meta_item.content {
+            Some(Loadable::Ready(meta_item)) => Some(Some(meta_item)),
+            Some(Loadable::Loading) => Some(None),
+            _ => None,
+        })
+        .flatten()
```
elpiel, line 222: can be simplified and remove the `.flatten()` by just removing the `Some(Some(..))` & `Some(None)`

elpiel (approved):
LGTM!

tymmesyde:
Tested with stremio-core-web on stremio-web and works, but it requires a change in core-web to add Clone derive on WebModel because the trait Model requires it
This also mean that the Clone derive is needed on all the core models that core-web uses

We could also check if `videos` is empty here so that we don't return a videos deeplink instead of checking for defaultVideoId ?
https://github.com/Stremio/stremio-core/blob/development/src/deep_links/mod.rs#L191 
And instead of overriding we could guess the id while we deserialize `MetaItemPreview` so we would have a correct streams deeplink ?

src/models/meta_details.rs
unclekingpin: Discussed in PM

unclekingpin:
@tymmesyde 
making changes on the deserialization logic will simplify things in the model but i didnt make it like that because:
1. we dont wrap the meta item with deep links anyway https://github.com/Stremio/stremio-core-web/blob/development/src/model/serialize_meta_details.rs#L83
2. in future we may need access to the original data for something else.

I will add Clone to all models later right after this pr. just wanted to make sure we are choosing this approach, i need @elpiel  to test the #471 and confirm or deny that those changes to the Env are actually needed. i tried to do it without env changes but couldnt.

elpiel:
Merge as you see fit.
The Clone introduction is not coupled with the bug, we can leave the Clone and remove the Model from the `RuntimeEvent` later on (it's only exposed to tests anyway)

PS: I'm looking into the bug.

PR #474, opened 2023-07-05
Override meta details selected

unclekingpin:
- derive Clone for all models
- do not override the MetaDetails Selected for meta_items with a single video


elpiel (approved):
LGTM!

PR #480, opened 2023-07-17
Continue watching part 2

elpiel:
Based on #450


src/models/continue_watching_preview.rs
```diff
@@ -1,44 +1,107 @@
...
 impl<E: Env + 'static> UpdateWithCtx<E> for ContinueWatchingPreview {
     fn update(&mut self, msg: &Msg, ctx: &Ctx) -> Effects {
+        // update the CW list if
         match msg {
-            Msg::Internal(Internal::LibraryChanged(_)) => {
-                library_items_update(&mut self.library_items, &ctx.library)
+            // library has changed
+            Msg::Internal(Internal::LibraryChanged(_))
+            // LibraryItem has been updated (this message alters the `mtime` and will re-order the CW list)
+            | Msg::Internal(Internal::UpdateLibraryItem(_))
```
elpiel, line 39: This might not be ideal, because it does update mtime, however, it does so even if you mark a movie or a video (episode) as watched.

- [x] Discuss with @jaruba whether this is intended consequence

PR #482, opened 2023-07-20
refactor(action): remove camelCase serde derive

tymmesyde:
To keep consistency with other actions

### DIFF
--- a/src/runtime/msg/action.rs
+++ b/src/runtime/msg/action.rs
@@ -128,7 +128,7 @@ pub enum ActionLoad {
 }
 
 #[derive(Clone, Deserialize, Debug)]
-#[serde(tag = "action", content = "args", rename_all = "camelCase")]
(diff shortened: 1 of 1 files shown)


elpiel (changes requested):
Actually I think it's better to add `rename_all = "camelCase"` to the rest of the actions that don't have it.
It's ~only~ applied to fields and variants (in enums) and most of the fields of the structs used in actions do not have `_` to begin with but our consistent way is to use camelCase across core.

tymmesyde:
> Actually I think it's better to add `rename_all = "camelCase"` to the rest of the actions that don't have it. It's ~only~ applied to fields and variants (in enums) and most of the fields of the structs used in actions do not have `_` to begin with but our consistent way is to use camelCase across core.

That means that we will need to rename all of our action names in our apps

elpiel (approved):
Ok, let's keep it snake_case.

PR #486, opened 2023-08-01
feat: toggle notifications for library item

tymmesyde:


src/models/ctx/update_library.rs
```diff
@@ -78,6 +78,25 @@ pub fn update_library<E: Env + 'static>(
             }))
             .unchanged(),
         },
+        Msg::Action(Action::Ctx(ActionCtx::ToggleLibraryItemNotifications(id))) => {
```
elpiel: - [x] Makes more sense to send a bool with the action to indicate the Subscribing or Unsubscrbing.

tymmesyde: Why ? This way we don't have to pass the state that the core already know
It will be a toggle/checkbox

elpiel: It's better to send specific state, UI can use toggle functionality but for backends it's better to set the state explicitly.
This removes the cases where you accidentally have 2 requests/actions and you flip the state back to its original state instead of `true`

It's also consistent with other actions.

tymmesyde: The core is synchronous so this should not happen

elpiel (approved):

elpiel:
This PR also reverts the renaming of the `noNotif` field since this requires changes in API which will cause a lot of problems.

PR #490, opened 2023-08-04
chore(docs): Document StreamDeepLinks & ExternalPlayerLink From impls

elpiel:
Remaining Questions :
- ~in `core-web` we use the StreamingServer base_url (if it has loaded it) to create the `StreamDeepLinks`, however, in other places we actually use the `Settings.streaming_server_url`. Should we default to the Settings or should we keep a way to override the url when creating StreamDeepLinks with the From impls?~ We should always use the `StreamingServer::base_url`.

https://github.com/Stremio/stremio-core-web/blob/8ff25c12cf507cf5e1e9ca080c1f6195c088d909/src/model/serialize_discover.rs#L186-L190

https://github.com/Stremio/stremio-core-web/blob/8ff25c12cf507cf5e1e9ca080c1f6195c088d909/src/model/serialize_discover.rs#L104-L107

### DIFF
--- a/src/deep_links/mod.rs
+++ b/src/deep_links/mod.rs
@@ -1,18 +1,25 @@
-mod error_link;
-
-use crate::constants::URI_COMPONENT_ENCODE_SET;
-use crate::deep_links::error_link::ErrorLink;
-use crate::models::installed_addons_with_filters::InstalledAddonsRequest;
-use crate::models::library_with_filters::LibraryRequest;
-use crate::types::addon::{ExtraValue, ResourcePath, ResourceRequest};
-use crate::types::library::LibraryItem;
-use crate::types::profile::Settings;
-use crate::types::query_params_encode;
-use crate::types::resource::{MetaItem, MetaItemPreview, Stream, StreamSource, Video};
 use percent_encoding::utf8_percent_encode;
 use regex::Regex;
 use serde::Serialize;
 use url::Url;
+
+use crate::{
+    constants::URI_COMPONENT_ENCODE_SET,
+    models::{
+        installed_addons_with_filters::InstalledAddonsRequest, library_with_filters::LibraryRequest,
+    },
+    types::{
+        addon::{ExtraValue, ResourcePath, ResourceRequest},
+        library::LibraryItem,
+        profile::Settings,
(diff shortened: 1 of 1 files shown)


src/deep_links/mod.rs
```diff
@@ -41,6 +41,86 @@ pub struct ExternalPlayerLink {
...
+            open_player,
+            android_tv,
+            tizen,
+            webos,
+            file_name,
+        }
+    }
+}
+
 impl From<(&Stream, &Option<Url>, &Settings)> for ExternalPlayerLink {
```
elpiel, line 51: - [ ] Should be removed if we don't want to be able to override the `streaming_url`

src/deep_links/mod.rs
```diff
@@ -300,6 +380,23 @@ pub struct StreamDeepLinks {
...
+                        utf8_percent_encode(&stream, URI_COMPONENT_ENCODE_SET),
+                    )
+                })
+                .unwrap_or_else(|error| ErrorLink::from(error).into()),
+            external_player: ExternalPlayerLink::from((stream, settings)),
+        }
+    }
+}
+
 impl From<(&Stream, &Option<Url>, &Settings)> for StreamDeepLinks {
```
elpiel, line 316: - [ ] Should be removed if we don't want to be able to override the `streaming_url`

tymmesyde:
We shouldn't do this because we are passing the `base_url` field of the `streaming_server` as `streaming_server_url` which is also used to tell if the server is online, using the `streaming_server_url` field from the `settings` would mean that it will be always treated as if it is online

elpiel:
> We shouldn't do this because we are passing the `base_url` field of the `streaming_server` as `streaming_server_url` which is also used to tell if the server is online, using the `streaming_server_url` field from the `settings` would mean that it will be always treated as if it is online

I have a recollection that we use the Settings streaming_server_url somewhere but it seems that it's a bad recollection in my memory. Indeed we always use the StreamingServer model to pass the url.
I will revert the changes and make sure to add comments what we need to pass to the `From` impl.

elpiel:
@tymmesyde has approved the changes in DM.
Merging...

PR #495, opened 2023-08-08
feat: update Toggle notif. for LibraryItem

elpiel:


tymmesyde (changes requested):
`state` doesn't need to be renamed to `no_notifications` as the action says it's a toggle 
No need to rename the fields of the action to camelCase as it's serialized from an array

elpiel:
> `state` doesn't need to be renamed to `no_notifications` as the action says it's a toggle No need to rename the fields of the action to camelCase as it's serialized from an array

Change has been reverted.

tymmesyde (approved):

PR #501, opened 2023-08-14
feat: use last_video_released for notifications

elpiel:
- player - update on switch_to_next_video
- notifs - update on dismiss and pull

# TODO:
- [ ] player - mark video as watched
- [ ] player - mark as watched


elpiel:
We won't be using this field but we will keep the logic with last_watched

PR #502, opened 2023-08-15
Set Default Value for profile.player_type

jaruba:


jaruba:
closed in favor of: https://github.com/Stremio/stremio-web/pull/443

PR #527, opened 2023-09-26
Resolve suggested stream in meta details model

TheBeastLT:
This PR adds a new field to meta details model - `suggested_stream`. This field finds a stream from addon responses, which should be highlighted or played if binge watching.
The logic is that we find the latest `StreamItem` stored based on last 30 videos from current video (ie. we're in E4, so we're going to check E4, E3, E2, E1 in this order until we hit a stored `StreamItem`). Then with the stream item we try to find a stream from addon responses (including the streams inside the meta itself `meta_streams`) - we find the responses from the addon based on `StreamItem.stream_transport_url`, then first we try to find the stream based on equality (as otherwise stored stream might be expired/no longer valid), if not found we try to find a stream based on it's `bingeGroup`. 
One note, why we cannot return `StreamItem.stream` directly if it's for the same episode, is that user might have played a stream from an addon which he no longer has due to some constrains (ie p2p addon), that's why we have to try to find it first and verify that's it's still available.


src/models/meta_details.rs
```diff
@@ -355,6 +383,74 @@ fn streams_update<E: Env + 'static>(
...
+                                        video_id: video.id.to_owned(),
+                                    })
+                                })
+                        })
+                })
+                .and_then(|stream_item| {
+                    [meta_streams, streams]
+                        .concat()
+                        .iter()
+                        .find(|resource| resource.request.base == stream_item.stream_transport_url)
```
elpiel, line 481: Do we expect only 1 found match? If not then maybe instead of using `find` we could find with the match for the `resource.content` merging the two lines in order to find a `Loadable::Ready(streams)` using the `find` all

src/models/meta_details.rs
```diff
@@ -355,6 +383,74 @@ fn streams_update<E: Env + 'static>(
     }
 }
 
+fn suggested_stream_update(
```
elpiel, line 440: Could you at least document the function so that we know what it's suppose to do in case of a bug, we could track that logic (intended vs actual)

elpiel (commented):
LGTM so far, I need to run some tests - a suitable test case scenario would be appreciated.

Just to confirm what is expected with this change:
In a nutshell, suggested stream tries to find a StreamItem for the past 30 episodes and if it find one it tries to find the same Base url (or simply put, stream from the same Addon) in the `meta_streams` & `streams` of the MetaDetails.

Can you briefly share what is the usecase for this suggested stream?

src/models/meta_details.rs
TheBeastLT: Addon url should be unique, as you cannot install multiple addons with same url.

src/models/meta_details.rs
TheBeastLT: What format would you suggest for this. as other methods are not documented?

elpiel: A lot of the information for this method was provided in the PR comment. Use that and include that at this point we're talking about the `Next video id` while binge watching.

src/models/meta_details.rs
elpiel: ahhh that's addon URL not resource URL. Resolved then.

src/models/meta_details.rs
elpiel: Small tweak I've made is to use the actual Doc comments `///` that will show on the documentation instead of regular comments `//`

elpiel (commented):
Code looks good to me, I will make sure to figure out a test-case to test it.

elpiel (approved):
Tested and works on web.
When there's an older episode with different stream - it chooses it.
For new and older episodes which are not watched it chooses (given the right binge watching ground and a stream for a full season) the correct suggested stream

PR #528, opened 2023-09-27
Resolve stream proxied streaming link when url has request headers

TheBeastLT:
Also streaming server now support `fileIndex=-1` when no index is available and will resolved the biggest file.

### DIFF
--- a/src/types/resource/stream.rs
+++ b/src/types/resource/stream.rs
@@ -14,7 +14,7 @@ use serde_with::{serde_as, DefaultOnNull};
 use std::collections::HashMap;
 use std::io::Write;
 use stremio_serde_hex::{SerHex, Strict};
-use url::Url;
+use url::{form_urlencoded, Url};
 
 #[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Debug)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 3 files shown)


src/types/resource/stream.rs
```diff
@@ -136,7 +136,38 @@ impl Stream {
...
+                            response
+                                .iter()
+                                .map(|header| ("r", format!("{}:{}", header.0, header.1))),
+                        );
+                        streaming_url
+                            .path_segments_mut()
+                            .ok()?
+                            .push("proxy")
+                            .push(proxy_query.finish().as_str())
+                            .push(&url.path()[1..]);
```
elpiel, line 164: It would be nice to document this format. I suppose this is a format that the Streaming server defines?
It could be beneficial to make a trait and implement it on Url for building this streaming url with proxy.
1. it's much easier to document
2. It will make it easy to test later on as well as a unit and not part of the `Stream::streaming_url`

src/types/resource/stream.rs
```diff
@@ -149,9 +180,9 @@ impl Stream {
                 match url.path_segments_mut() {
                     Ok(mut path) => {
                         path.push(&hex::encode(info_hash));
-                        if let Some(file_idx) = file_idx {
-                            path.push(&file_idx.to_string());
-                        }
+                        path.push(
+                            &file_idx.map_or_else(|| "-1".to_string(), |idx| idx.to_string()),
+                        );
```
elpiel, line 187: Could you once more document when we use `-1` as value. As someone that's still relatively new to the project, I'm not aware of all these options and would be nice for future contributors to also have this knowledge in the source code

src/unit_tests/deep_links/stream_deep_links.rs
elpiel, line 1: Tests look ok!

elpiel (commented):
My suggestions are purely for readability and documentation.

Please share a test scenario that we can try out this change with.

src/types/resource/stream.rs
TheBeastLT: SInce this is used only on specific conditions and is needed here only, I personally wouldn't make it as a trait and leave it here, where all other streaming links are built.

elpiel: ok, I just wanted to isolate it instead of testing with a full Stream object.

elpiel (approved):
LGTM and tested using custom Addon. Returned url seems encoded and correct:

http://192.168.0.3:11470/proxy/d=https%253A%252F%252Fdownload.blender.org&h=req-header-1%253Avalue-1&r=resp-header-1%253Avalue-1/peach%2Fbigbuckbunny_movies%2Fbig_buck_bunny_720p_h264.mov

PR #536, opened 2023-10-12
feat: player next stream

tymmesyde:


src/unit_tests/player/next_stream.rs
```diff
@@ -0,0 +1,158 @@
...
+fn next_stream() {
+    #[derive(Model, Default, Clone, Debug)]
+    #[model(TestEnv)]
+    struct TestModel {
+        ctx: Ctx,
+        player: Player,
+    }
+
+    fn fetch_handler(request: Request) -> TryEnvFuture<Box<dyn Any + Send>> {
+        println!("{:?}", request.url);
```
elpiel: you can remove this `println`

elpiel (commented):
Does this PR remove the need to go through the MetaDetails model before loading the next (suggested) stream (i.e. #527) ?

Basically the same suggested stream should be returned from the Player itself I suppose.

This leads me to another question: What about the StreamsBucket? Shouldn't we update the Player model to also takes care of adding the Stream to the StreamsBucket?
Right now on `Player::Load` it adds the selected stream to the StreamBucket but I'm not sure if we should do this beforehand when selecting `next_stream`.

tymmesyde:
> Does this PR remove the need to go through the MetaDetails model before loading the next (suggested) stream (i.e. #527) ?

I'm not sure, we should discuss it

> Basically the same suggested stream should be returned from the Player itself I suppose.

Right, either we return the `next_stream` object in our different implementations of core or use it to build deep links for `next_video`

> This leads me to another question: What about the StreamsBucket? Shouldn't we update the Player model to also takes care of adding the Stream to the StreamsBucket? Right now on `Player::Load` it adds the selected stream to the StreamBucket but I'm not sure if we should do this beforehand when selecting `next_stream`.

I'm not sure if it's an issue

elpiel (approved):
LGTM! Let's discuss this with @TheBeastLT in Slack and see if the other applications that go through the MetaDetails first to keep the history navigation correct should remain or we should use the `next_stream` of this PR

elpiel:
Related #397
- [x] ~We should also add the bindge watch group when loading the `next_streams`~ It's better to keep them all and filter accordingly to out needs

PR #540, opened 2023-10-20
feat: MetaDetails - guess streams only when a flag is present

elpiel:
This feature allows you to load the MeteDetails without loading streams.
Previous feature allowed the guessing of the streams if a stream_path was not provided - see #474, but there are cases in the apps where we load the `MetaDetails` but we don't actually want to load the streams as we only show a small modal with just a few things from the MetaDetails like the Trailer.

We now have 4 code cases:
1. `stream_path: Some` and we have `guess_stream: false` - loads the stream from the provided path
2. `stream_path: Some` and `guess_stream: true` - loads the stream from the provided path ignoring the `guess_stream: true`
3. `stream_path: None` and `guess_stream: true` - guesses the `stream_path` using the first successfully loaded `MetaItem`
4. `stream_path: None` and `guess_stream: false` - no streams will be loaded for the `MetaDetails`

### DIFF
--- a/src/models/meta_details.rs
+++ b/src/models/meta_details.rs
@@ -32,7 +32,9 @@ pub struct Selected {
     pub meta_path: ResourcePath,
     pub stream_path: Option<ResourcePath>,
     #[serde(default)]
-    pub stream_override: bool,
+    /// if `stream_path` is `None` then we try to guess the video and make a request
+    /// to the addons to load the streams for that video id
+    pub guess_stream: bool,
 }
 
 #[derive(Default, Serialize, Clone, Debug)]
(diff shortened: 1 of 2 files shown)


src/models/meta_details.rs
```diff
@@ -255,6 +282,9 @@ fn selected_override_update(
                 id: video_id,
                 extra: vec![],
             }),
+            // we must set the stream_override to false after we've overridden it
+            // to make it consistent
+            guess_stream: false,
```
TheBeastLT, line 287: I guess it's not needed to be overriden now that you check if it's none.

TheBeastLT (approved):

src/models/meta_details.rs
elpiel: yes, that's why it's `false`

TheBeastLT: No, I meant that this line is no longer needed, as we are setting the stream_path, and this function won't trigger anymore anyway.

elpiel: You need to set a value for the field, `false` is the default + you don't want to guess the stream path when we've overridden it

tymmesyde (approved):

tymmesyde:
Maybe `guess_stream` true by default since it's the intended behavior ?

elpiel:
> Maybe `guess_stream` true by default since it's the intended behavior ?

no, it's not. The value `true` should be taken into account **only** when `stream_path` is `None`

PR #541, opened 2023-10-20
migrate to storage schema 8

unclekingpin:


elpiel (approved):

elpiel:
Needs tests for the migration but we can add it later.

PR #544, opened 2023-10-22
remove Default impl from Ctx

unclekingpin:


elpiel (commented):
Would it be better to add PhantomData with a generic Env. and use E::now? Not sure if this would actually work with the derivative attr. macro but we could try?!

elpiel (approved):

PR #546, opened 2023-10-24
feat: Profile Settings - controls & esc for exiting fullscreen

elpiel:
- Adds a new field `seek_short_time_duration` for short (finer) seeking (e.g. using Arrow + Shift) and with default value of - **3 seconds**
- Adds setting for `Escape key exiting Full screen` of the app
- Adds setting for `Pause playback on minimize`
Discussed on Slack: Since most players have default seeking time of **10 seconds** we will keep this default and instead provide a finer seek time of **3 seconds**.

Related issue: https://github.com/Stremio/stremio-web/issues/413


src/types/profile/settings.rs
```diff
@@ -26,7 +26,12 @@ pub struct Settings {
     pub subtitles_text_color: String,
     pub subtitles_background_color: String,
     pub subtitles_outline_color: String,
+    /// Whether or not the Escape key should exists from the app when in Full screen.
+    pub esc_exists_fullscreen: bool,
```
tymmesyde: Should be `esc_exit_fullscreen`

tymmesyde (changes requested):

src/types/profile/settings.rs
```diff
@@ -26,7 +26,12 @@ pub struct Settings {
...
+    /// The Seek time duration (in milliseconds) is when using the Arrow keys
     pub seek_time_duration: u32,
+    /// The Seek shift time duration (in milliseconds) is when using the Arrow keys + Shift
+    pub seek_shift_time_duration: u32,
```
TheBeastLT: Instead of changing logic for the whole variable and migrating it, wouldn't it be better for the new variable to be different ie. `seek_long_time_duration`. As it will create an easy error, where the old variable name stays the same, but it's value is changed. Since `seek_time_duration` was meant to be used as the default seek time with the default method of seeking. 
Now it's just hardcoding logic in settings which is meant for web only. (there's no shift on mobile or tablet or tvs). And then IMO it should be a completely new setting for web only, which doesn't impact any other services.

elpiel: The old variable is the long one, using the Shift key is the finer seeking.
I'm open for a new name.

elpiel: This means that we still need to migrate the value as that's the default seeking that we want to change in a sense.

elpiel: as discussed on slack, I've update the PR and initial comment to reflect the discussion.
I'm resolving this conversation.

src/types/profile/settings.rs
elpiel: Fixed.

tymmesyde: It should be `esc_exit_fullscreen` not `esc_exist_fullscreen`

TheBeastLT (approved):

src/types/profile/settings.rs
elpiel: lol, fixed the typo,again :D

tymmesyde (approved):

PR #548, opened 2023-10-26
Local search was loading upon instantiation

tymmesyde:
Fix https://github.com/Stremio/stremio-tasks/issues/44

Was loading upon instantiation instead of load action, causing theater to freeze on startup


elpiel (approved):
Makes perfect sense. I thought that these requests in the Env are async?!
We definitely need to take a look at the Env again and fix it once and for all...

PR #551, opened 2023-11-01
Fix merge conflict with development

elpiel:


src/models/meta_details.rs
```diff
@@ -73,7 +86,6 @@ impl<E: Env + 'static> UpdateWithCtx<E> for MetaDetails {
                     .join(meta_items_effects)
                     .join(meta_streams_effects)
                     .join(streams_effects)
-                    .join(suggested_stream_effects)
```
TheBeastLT: probably this should stay here?

elpiel: Fixed! Thanks for noticing.

Issue #55, opened 2019-06-06
Player model


Ivshti:
```rust
pub struct PlayerModel {
   // from the router, we will get (item_id, video_id, stream, meta_resourcereq)
   // @TODO maybe an addon transport_url too, there are use cases where we want information from the addon manifest
   // @TODO: Option or Loadable? Cause we may want to express "will not load" vs "loading"
   pub meta: Option<MetaItem>,
   pub next_video: Option<Video>,
   pub libitem: Option<LibItem>, // should we use Loadable? 
   pub stream: Stream,
}
```

As for handling the bingeing (or other complicated cases), we should:

```rust
fn update(&mut self, ctx: &Ctx, action: &Action) {
    match action {
        Msg::Player(item, PlayerEv::Ended) if item == Some(self.libitem) => {
              if let Some(next) = self.next_video && ctx.settings.bingeing  {
                    // trigger an Effect that resolves in Open
              }
        }

```

By default, the player will try to get an existing item from the library with this ID. If it does not exist, it will construct it from the loaded meta. If that doesn't load, it will operate without a libItem.

All extra roles: bingeing, saving library item state, marking episodes watched, marking notifications seen

It can operate without a libItem (just proxying the Player implementation) or with one, in which case it should  load the library item, nad mutate it as the player advances

When the player starts playing it should request subtitles from the addon system and do `addExtraSubtitles` for each result and perhaps set selected subtitles when applying PlayerPreferences; 
also send `stream.subtitles` via `addExtraSubtitles`

It should also mutate and persist PlayerPreferences itself, which should contain preferred subtitle language, preferred track for each language, preferred audio track, and subtitles delay

It should also catch and handle internal inconsistencies, most notably the playing `stream` being different than `self.stream`

nklhtv:
Some part of the logic is implemented: using the correct library item, binge watching, updating library item state.
The model does not act as a proxy of the player impl as we cannot keep references to ui elements or js objects in core. Instead the ui loads the stream thats currently exported by the player model, and transforms various player impl events into player actions: onEnded, onTimeChange etc.

Player model still needs to:

- load subtitles using the video hash provided by the player impl (if supported)
- manage user preferences about subtitles and audio tracks

elpiel:
Video params have been added/changed in #533 
Subtitles and audio tracks state tracking has been added in #553 

User settings for the audio track language and subtitles language are handled by each application and not in core.
A stream override of these audio and subtitles languages is now implemented in #553

PR #553, opened 2023-11-04
implement stream state management

TheBeastLT:
Allow storing stream state based on user actions (user changes default audio/subs track, changes sync or playback speed) and restoring it when playing the same stream or reusing parts of the state when playing next videos in the meta.


src/types/streams/streams_item.rs
```diff
@@ -15,7 +15,60 @@ pub struct StreamsItem {
...
     pub mtime: DateTime<Utc>,
 }
+
+#[serde_as]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct StreamItemState {
+    pub subtitle_track_id: Option<String>,
+    pub subtitle_language: Option<String>,
+    pub subtitle_delay_ms: Option<i64>,
```
elpiel: I would prefer the `subtitles_delay` and `audio_delay` to be documented to being milliseconds instead of adding it to the name + you can usually make a struct to specify that it's a delay of milliseconds if you want to be that strict.

It will be consistent with other parts of core where we have such values

TheBeastLT: Do we have such a struct for miliseconds?

elpiel:
Closes #55

src/types/streams/streams_item.rs
elpiel: Not right now, that's why I suggest just removing the MS from the field name

src/models/ctx/update_streams.rs
```diff
@@ -42,12 +46,38 @@ pub fn update_streams<E: Env + 'static>(
...
+            };
+            let steam_item = streams.items.get(&key).cloned();
+            match steam_item {
+                Some(item) => {
+                    let new_stream_item = StreamsItem {
+                        state: Some(state.to_owned()),
+                        ..item
+                    };
+                    streams.items.insert(key, new_stream_item);
+                    Effects::msg(Msg::Internal(Internal::StreamsChanged(false)))
```
elpiel: This job will be best fitted for `entry` as it gives you the option to override the entry if it's taken instead of replacing it with `insert` as it is now.

src/types/streams/streams_item.rs
```diff
@@ -15,7 +15,65 @@ pub struct StreamsItem {
...
+    pub state: Option<StreamItemState>,
     /// Modification time
     #[serde(rename = "_mtime")]
     pub mtime: DateTime<Utc>,
 }
+
+/// StreamItemState is to be used when user intentionally changes some values from the defaults,
+/// so that they would be persisted and restored when returning to the same stream,
+/// or some of them reapplied when moving to the next video/stream.
+#[serde_as]
```
elpiel: Not needed

tymmesyde:
What are `subtitle_language`, `audio_language` and `player_type` for ?

tymmesyde:
Should probably be named to reflect the stremio-video properties names:
https://github.com/Stremio/stremio-video/blob/master/src/HTMLVideo/HTMLVideo.js#L673

`subtitle_id` -> `selected_subtitles_track_id`
`audio_id` -> `selected_audio_track_id`
`subtitle_delay` -> `subtitles_offset`
etc...

tymmesyde:
Also do you think handling the extra subtitles tracks would be nice ?
Like `selected_extra_subtitles_track_id` to also set selected subtitles from addon requests

elpiel (commented):
2 small nitpicks. I need to test the PR more thoroughly.

src/models/ctx/update_streams.rs
TheBeastLT: Not sure I got you meaning, this was done the same when inserting stream item above. We need to store the changed state into storage so we need to launch this effect.

TheBeastLT:
> Also do you think handling the extra subtitles tracks would be nice ? Like `selected_extra_subtitles_track_id` to also set selected subtitles from addon requests

This is to store any state of any action the user does manually. It should not store any info that is selected by default based on settings.
`subtitle_track_id` can be either the id of the embedded subs or identifier of addon returned and manually selected subtitle.
`language` param for audio and subtitles is required if you want to maintain the same embedded track selected when moving to the next video. Their `id` is usually the index of the track in video file, but since track order might change then it's best to also have the language so that you would be sure that in the next video you're selecting the correct track (id and lang must match).
`player_type` is required since in android tv you can change the player on demand from the player view, so you'd want to maintain this info and use the same player throughtout the series. (Ie. on TV some users prefer using libVLC on anime videos due to better ass subs support)

tymmesyde:
> > Also do you think handling the extra subtitles tracks would be nice ? Like `selected_extra_subtitles_track_id` to also set selected subtitles from addon requests
> 
> This is to store any state of any action the user does manually. It should not store any info that is selected by default based on settings. `subtitle_track_id` can be either the id of the embedded subs or identifier of addon returned and manually selected subtitle. `language` param for audio and subtitles is required if you want to maintain the same embedded track selected when moving to the next video. Their `id` is usually the index of the track in video file, but since track order might change then it's best to also have the language so that you would be sure that in the next video you're selecting the correct track (id and lang must match). `player_type` is required since in android tv you can change the player on demand from the player view, so you'd want to maintain this info and use the same player throughtout the series. (Ie. on TV some users prefer using libVLC on anime videos due to better ass subs support)

In web / theater since we use `stremio-video`, it makes the distinction between embedded subtitles tracks and the ones provided by addons
`selectedSubtitlesTrackId` is for embedded tracks
`selectedExtraSubtitlesTrackId` is for addon tracks

For embedded tracks it's their index but for extra tracks it's the id returned by the addon:
https://github.com/Stremio/stremio-addon-sdk/blob/master/docs/api/responses/subtitles.md#subtitles-object

So maybe we can remove `*_language` and / or add `selected_extra_subtitles_track_id`,
This way we don't have to match for id (index) + language, we can just match the id

TheBeastLT:
> In web / theater since we use `stremio-video`, it makes the distinction between embedded subtitles tracks and the ones provided by addons `selectedSubtitlesTrackId` is for embedded tracks `selectedExtraSubtitlesTrackId` is for addon tracks
> 

You can only have one selected subtitle at the time, so there shouldn't be any distinction. What you have is web specific, but still you have to decide which subs you're displaying so in the end there should be a single active id.

> For embedded tracks it's their index but for extra tracks it's the id returned by the addon: https://github.com/Stremio/stremio-addon-sdk/blob/master/docs/api/responses/subtitles.md#subtitles-object
> 
> So maybe we can remove `*_language` and / or add `selected_extra_subtitles_track_id`, This way we don't have to match for id (index) + language, we can just match the id

Like I mentioned, since it's an index for embedded tracks, you need some extra protection to check that if you want to select that track for the next stream that the order didn't change, so we need the language.

tymmesyde:
> > In web / theater since we use `stremio-video`, it makes the distinction between embedded subtitles tracks and the ones provided by addons `selectedSubtitlesTrackId` is for embedded tracks `selectedExtraSubtitlesTrackId` is for addon tracks
> 
> You can only have one selected subtitle at the time, so there shouldn't be any distinction. What you have is web specific, but still you have to decide which subs you're displaying so in the end there should be a single id.
> 
> > For embedded tracks it's their index but for extra tracks it's the id returned by the addon: https://github.com/Stremio/stremio-addon-sdk/blob/master/docs/api/responses/subtitles.md#subtitles-object
> > So maybe we can remove `*_language` and / or add `selected_extra_subtitles_track_id`, This way we don't have to match for id (index) + language, we can just match the id
> 
> Like I mentioned, since it's an index for embedded tracks, you need some extra protection to check that if you want to select that track for the next stream that the order didn't change, so we need the language.

I mean we **need** `selected_extra_subtitles_track_id` for every apps that uses [stremio-video](https://github.com/Stremio/stremio-video)
Otherwise we don't know if `subtitle_track_id` is for embedded or extra / addon tracks

TheBeastLT:
> > > In web / theater since we use `stremio-video`, it makes the distinction between embedded subtitles tracks and the ones provided by addons `selectedSubtitlesTrackId` is for embedded tracks `selectedExtraSubtitlesTrackId` is for addon tracks
> > 
> > 
> > You can only have one selected subtitle at the time, so there shouldn't be any distinction. What you have is web specific, but still you have to decide which subs you're displaying so in the end there should be a single id.
> > > For embedded tracks it's their index but for extra tracks it's the id returned by the addon: https://github.com/Stremio/stremio-addon-sdk/blob/master/docs/api/responses/subtitles.md#subtitles-object
> > > So maybe we can remove `*_language` and / or add `selected_extra_subtitles_track_id`, This way we don't have to match for id (index) + language, we can just match the id
> > 
> > 
> > Like I mentioned, since it's an index for embedded tracks, you need some extra protection to check that if you want to select that track for the next stream that the order didn't change, so we need the language.
> 
> I mean we **need** `selected_extra_subtitles_track_id` for every apps that uses [stremio-video](https://github.com/Stremio/stremio-video) Otherwise we don't know if `subtitle_id` is for embedded or extra / addon tracks

Atm in core subtitle id should be the url to the subtitle itself as that's unique, so it will be very different from embedded subs id, so you will be able to just iterate through combined array of embedded subs and addons subs and determine which ones need to be selected.

tymmesyde:
> Atm in core subtitle id should be the url to the subtitle itself as that's unique, so it will be very different from embedded subs id, so you will be able to just iterate through combined array of embedded subs and addons subs and determine which ones need to be selected.

I'm talking about `subtitle_track_id` from this PR, we need a way to know if this is an id for embedded track or extra subtitles track
So i suggest adding another field `extra_subtitles_track_id`

TheBeastLT:
> I'm talking about `subtitle_track_id` from this PR, we need a way to know if this is an id for embedded track or extra subtitles track So i suggest adding another field `extra_subtitles_track_id`

Don't think we're on the same page here, so to clarify:
`subtitle_track_id` in `StreamState` is an identifier of currently shown subtitle the user manually selected (subtitle chosen by default based on preference settings shouldn't be stored here). There can only be one subtitles shown at a time regardless whether it's embedded or addon subtitle.

Embedded subtitles usually come in format `(index,name,lang)` and for them `index` would be the unique identifier which is a number.

Addon subtitle response in core has structure `(lang,url)` where unique identifier for them would be url.

So at any point you can always determine the subtitle `subtitle_track_id` based on these unique identifiers and you'll either find the sub in embedded track when and index was stored (together checking the lang as well to be safe) or iterating subtitle responses and checking based on url.

Having `subtitle_track_id` and then `extra_subtitle_track_id` will lead to having corrupt state, since when you populate both of them you won't know which the user actually choose, the embedded one or the addon one, so there needs to be a single id stored. 
Same is done currently in desktop/mobile, just that it's stored per language preference https://github.com/Stremio/stremio-models/blob/baa9ac813c2fcb2830a6af7110110a56edb24135/subtitles.js#L482

src/models/ctx/update_streams.rs
elpiel: instead of `get` and later `insert` you can use `entry()` and handle the case where it's `Occupied` entry and directly mutate the value in the HashMap:

https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html

It's a more idiomatic way of doing what you've done here. Not related to the business logic itself

src/types/streams/streams_bucket.rs
```diff
@@ -28,4 +29,32 @@ impl StreamsBucket {
             items: HashMap::new(),
         }
     }
+
+    pub fn last_stream_item(
+        &self,
+        video_id: &String,
```
elpiel:
```suggestion
        video_id: &str,
```

src/types/streams/streams_bucket.rs
```diff
@@ -28,4 +29,32 @@ impl StreamsBucket {
...
+    ) -> Option<&StreamsItem> {
+        match meta_item.videos.len() {
+            0 => self.items.get(&StreamsItemKey {
+                meta_id: meta_item.preview.id.to_string(),
+                video_id: video_id.to_owned(),
+            }),
+            _ => meta_item
+                .videos
+                .iter()
+                .position(|video| video.id == *video_id)
```
elpiel:
```suggestion
                .position(|video| &video.id == video_id)
```

src/types/streams/streams_bucket.rs
```diff
@@ -28,4 +29,32 @@ impl StreamsBucket {
...
+
+    pub fn last_stream_item(
+        &self,
+        video_id: &String,
+        meta_item: &MetaItem,
+    ) -> Option<&StreamsItem> {
+        match meta_item.videos.len() {
+            0 => self.items.get(&StreamsItemKey {
+                meta_id: meta_item.preview.id.to_string(),
+                video_id: video_id.to_owned(),
```
elpiel, line 37:
```suggestion
                video_id: video_id.to_string(),
```

src/types/streams/streams_item.rs
```diff
@@ -15,7 +15,98 @@ pub struct StreamsItem {
...
+    pub subtitle_track: Option<SubtitleTrack>,
+    pub audio_track: Option<AudioTrack>,
+    pub playback_speed: Option<f32>,
+    pub player_type: Option<String>,
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SubtitleTrack {
+    /// Id of the subtitle track
```
elpiel, line 49:
```suggestion
    /// Id of the subtitle track - an index for embedded subtitles and an id for external ones.
```

src/types/streams/streams_item.rs
elpiel, line 1: It would be nice to include `#[serde(skip_serializing_if = "Option::is_none")]` to all fields which are `Option`-al since it will lower the amount of JSON we need to move through WASM, especially because users don't always override these settings for a stream

src/types/streams/streams_item.rs
TheBeastLT: Since this is not a given and depend on player implementations, I think it's better not to include it.

elpiel (approved):
LGMT, Great job!

PR #556, opened 2023-11-09
wrap suggested stream in resource loadable

TheBeastLT:
Wrap `suggested_stream` in a `ResourceLoadable` as it is valuable to know at what stage the suggested stream resolving is. Ie it's good to know that the required addon has finished loading, but no suggested stream was found, so then if you have some waiting screen you can stop waiting and display stream selection screen instead.


elpiel (approved):

elpiel:
LGTM!

PR #557, opened 2023-11-10
Fix external player link streaming server url

### LINKED ISSUE #404: Video url always using local server IP
Currently on web.stremio it's always using the local server IP when trying to fetch videos. 
If I added an external server url in the settings, then attempt to play a video it uses the local server IP as a hostname not the external server's which cause it to show an error saying I must use an external player and when I click that it downloads a .m3u file with an url with the local IP:
image
Meanwhile it's actually working correctly on the desktop app and the old app.strem.io website.

### LINKED ISSUE #431: Streaming URL hostname is wrong
Copying the streaming URL yields something like:

```http://10.0.0.140:11470/...``` ... which does not work, rather than using the hostname provided in streaming URL in settings (which does work)

When opening the stream in external player, it also does not work.

### PR DESCRIPTION
**Pull request by tymmesyde**
Close #539 
Fix https://github.com/Stremio/stremio-web/issues/404
Fix https://github.com/Stremio/stremio-web/issues/431

### DIFF
--- a/src/deep_links/mod.rs
+++ b/src/deep_links/mod.rs
@@ -64,6 +64,7 @@ impl From<(&Stream, Option<&Url>, &Settings)> for ExternalPlayerLink {
     ///
     /// [`StreamingServer::base_url`]: crate::models::streaming_server::StreamingServer::base_url
     fn from((stream, streaming_server_url, settings): (&Stream, Option<&Url>, &Settings)) -> Self {
+        // Use streaming_server_url from settings if streaming_server is reachable
         let streaming_server_url = streaming_server_url.map(|_| &settings.streaming_server_url);
         let http_regex = Regex::new(r"https?://").unwrap();
         let download = stream.download_url();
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by tymmesyde** [comment]
@edde746 Can you test this when you have the time and let me know if it's working for you ?

**Comment by edde746** [comment]
Works

**Comment by tymmesyde** [comment]
> Works

Thank you !

File: src/deep_links/mod.rs [hunk 1]
```diff
@@ -64,6 +64,7 @@ impl From<(&Stream, Option<&Url>, &Settings)> for ExternalPlayerLink {
     ///
     /// [`StreamingServer::base_url`]: crate::models::streaming_server::StreamingServer::base_url
     fn from((stream, streaming_server_url, settings): (&Stream, Option<&Url>, &Settings)) -> Self {
+        let streaming_server_url = streaming_server_url.map(|_| &settings.streaming_server_url);
```
elpiel, line 68:
```suggestion
        // Only if the streaming server is available, make sure to use the settings url
        // as it might contain a remote url of the server,
        // otherwise, deeplinks will be broken for remote server
        let streaming_server_url = streaming_server_url.map(|_| &settings.streaming_server_url);
```

elpiel (approved):

PR #561, opened 2023-11-13
feat: Seek log

elpiel:
- Adds a new PlayerAction for playing the `NextVideo`
- creates seek logs when the user seeks the video
- sends seek log request after `Unload` (with no outro information) or `NextVideo` action (with current offset as `outro`)


src/runtime/msg/action.rs
```diff
@@ -143,6 +143,13 @@ pub enum ActionPlayer {
     PausedChanged {
         paused: bool,
     },
+    /// Play next video, if there is one, applicable to e.g.
+    /// movie series and playing the next episode.
+    NextVideo,
+    /// Video player has ended.
+    /// 2 scenarios are possible:
+    /// - We've watched a movie to the last second
+    /// - We've watched a movie series to the last episode and the last second
```
TheBeastLT: Not sure if it indicates that we've watched to the last episode. This just indicates that the end of currently playing stream was reached.

elpiel: Good point, will check with @tymmesyde as well

src/models/player.rs
```diff
@@ -451,6 +495,30 @@ impl<E: Env + 'static> UpdateWithCtx<E> for Player {
                 };
                 trakt_event_effects.join(update_library_item_effects)
             }
+            Msg::Action(Action::Player(ActionPlayer::NextVideo)) => {
```
TheBeastLT, line 492: Since Unload action happens always as you leave the player view, probably it's a good idea that if we successfully submit seek history data on this next video action to clear  the history, so that we wouldn't submit it again on Unload, which will happen after this action.
Edit: ah, tho I see it's already done as so in the seek_update method 👍

TheBeastLT (approved):

src/models/player.rs
```diff
@@ -873,6 +940,68 @@ fn subtitles_update<E: Env + 'static>(
...
+) -> Effects {
+    let seek_request_effects = match (selected, video_params, series_info, library_item) {
+        (Some(selected), Some(video_params), Some(series_info), Some(library_item)) => {
+            match (
+                &selected.stream.source,
+                selected.stream.name.as_ref(),
+                video_params.hash.clone(),
+            ) {
+                (StreamSource::Torrent { .. }, Some(stream_name), Some(opensubtitles_hash)) => {
+                    let filename_hash = {
```
tymmesyde: Maybe it should be `st_hash` or `stream_name_hash` to not be confused with FileName:
https://github.com/Stremio/stremio-api/blob/master/rpc/seekLog.go#L50C2-L50C10

tymmesyde (approved):
Tested with web
Only thing that i noticed is that it sends the request even if the `seekHistory` and `skipOutro` are empty

src/models/player.rs
elpiel: Ok, will change

PR #562, opened 2023-11-14
feat: implement search history bucket

tymmesyde:
Related https://github.com/Stremio/stremio-tasks/issues/87

Implement search history bucket, updated when Load action for CatalogsWithExtra is dispatched with search extra param
Also add a Ctx action to clear the items of the bucket


src/types/search_history/search_history_bucket.rs
```diff
@@ -0,0 +1,23 @@
+use std::collections::HashMap;
+
+use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+
+use crate::types::profile::UID;
+
+#[serde_as]
```
elpiel: not needed.

src/unit_tests/ctx/update_search_history.rs
```diff
@@ -0,0 +1,152 @@
...
+            action: Action::Load(ActionLoad::CatalogsWithExtra(Selected {
+                r#type: None,
+                extra: vec![ExtraValue {
+                    name: "search".to_owned(),
+                    value: "superman".to_owned(),
+                }],
+            })),
+        })
+    });
+
```
elpiel, line 140: I would add one check here to make sure we actually have something in the search history just in case the action doesn't do what it suppose to do.

elpiel (approved):

PR #570, opened 2023-12-01
add `filename` video param

unclekingpin:


src/addon_transport/http_transport/legacy/mod.rs
```diff
@@ -213,6 +215,14 @@ fn build_legacy_req(transport_url: &Url, path: &ResourcePath) -> Result<Request<
             let video_size = path
                 .get_extra_first_value(VIDEO_SIZE_EXTRA_PROP.name.as_str())
                 .and_then(|video_size| video_size.parse().ok());
+            let video_filename =
```
TheBeastLT, line 224: Probably this should be below the `if let Some(video_size) = video_size {` block, to not split the logic blocks up.

src/models/player.rs
```diff
@@ -860,7 +862,8 @@ fn subtitles_update<E: Env + 'static>(
                         .extend_one(
                             &VIDEO_SIZE_EXTRA_PROP,
                             video_params.size.as_ref().map(|size| size.to_string()),
-                        ),
+                        )
+                        .extend_one(&VIDEO_FILENAME_EXTRA_PROP, video_params.filename.to_owned()),
```
TheBeastLT, line 866: Since this is going to be part of `Stream.behaviourHints` object, maybe we should take from there?

unclekingpin: that logic is in [stremio-video](https://github.com/Stremio/stremio-video/blob/master/src/withStreamingServer/fetchVideoParams.js#L4-L8)

src/addon_transport/http_transport/legacy/mod.rs
unclekingpin: updated

src/models/player.rs
TheBeastLT: Tho that means that every client will have to duplicate this logic, extracting values from `Stream.behaviourHints` and passing them via `VideoParams` instead of core taking care of that.

TheBeastLT (approved):

src/models/player.rs
unclekingpin: yes, every non js app :(
we can think of way to move that logic to rust

PR #571, opened 2023-12-01
prevent unnecessary library updates

unclekingpin:
prevent unnesessary library updates in local storage and api when only mtime is changed


elpiel (approved):
Lgtm

PR #575, opened 2023-12-05
fix: Player next_video - skip unreleased episodes

### LINKED ISSUE #574: Player's Next video loads episodes that haven't been released yet
If the next episode has already been added to our Meta Details then Player doesn't filter episodes that haven't been released when deciding on the next video to play in the player.

Screenshot from 2023-12-05 11-17-54

elpiel:
Fixes #574

### DIFF
--- a/src/models/player.rs
+++ b/src/models/player.rs
@@ -157,7 +157,7 @@ impl<E: Env + 'static> UpdateWithCtx<E> for Player {
                     &self.video_params,
                     &ctx.profile.addons,
                 );
-                let next_video_effects = next_video_update(
+                let next_video_effects = next_video_update::<E>(
                     &mut self.next_video,
                     &self.next_stream,
                     &self.selected,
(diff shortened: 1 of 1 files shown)


src/models/player.rs
```diff
@@ -664,12 +668,18 @@ fn next_video_update(
                     .unwrap_or_default();
                 next_season != 0 || current_season == next_season
             })
-            .map(|(_, next_video)| {
-                let mut next_video = next_video.clone();
-                if let Some(stream) = stream {
-                    next_video.streams = vec![stream.clone()];
+            .and_then(|(_, next_video)| {
+                match next_video.released {
```
TheBeastLT, line 672: Since this released property is not exactly accurate I'm not entirely sure whether it's a good idea not allow going to the next video based on it. Since I'd say it's best to see whether are are valid stream or not for the user. But don't have a strong opinion 🤷

TheBeastLT (approved):

elpiel:
after a discussion with @jaruba we are keeping the old behavior.
Also taking into account that the released date is not always accurate.

PR #576, opened 2023-12-05
fix: StreamingServer statistics response `null`

### LINKED ISSUE #573: Statistics - When null is returned from the server, it causes an error
When the stream has loaded to 100%, core still tries to deserialise the statistics object from the server request and ends up with an Error instead of keeping the last known Statistics

Ensure that we:
- [ ] Keep the last known statistics #577 
- [x] Handle `null` as "stream has loaded fully" #576

elpiel:
Resolves #573
Fix the issue where the server returns a `null` json response for the `stats.json`.
Up until now it was throwing an error on deserialization

### DIFF
--- a/src/models/streaming_server.rs
+++ b/src/models/streaming_server.rs
@@ -180,6 +180,7 @@ impl<E: Env + 'static> UpdateWithCtx<E> for StreamingServer {
                 let selected_effects =
                     eq_update(&mut self.selected.statistics, Some(request.to_owned()));
                 let statistics_effects = eq_update(&mut self.statistics, Some(Loadable::Loading));
+
                 Effects::one(get_torrent_statistics::<E>(
                     &self.selected.transport_url,
                     request,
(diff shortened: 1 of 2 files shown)


TheBeastLT (approved):

src/models/streaming_server.rs
```diff
@@ -588,7 +590,8 @@ fn get_torrent_statistics<E: Env + 'static>(url: &Url, request: &StatisticsReque
         .body(())
         .expect("request builder failed");
     EffectFuture::Concurrent(
-        E::fetch::<_, Statistics>(request)
+        // `null`` can be returned when the stream has been loaded 100%
```
tymmesyde: It's happening when the engine is destroyed for inactivity:
https://github.com/Stremio/enginefs/blob/8722f2ecf423fc98d529aff3d0db5864e501ba04/enginefs.js#L309
If it was downloaded to 100% and that the stream is paused, then played, it will create a new engine and return the correct stats

elpiel: Fixed the comment.

PR #580, opened 2023-12-08
 feat: Player - skip gaps API request

elpiel:


src/models/player.rs
```diff
@@ -197,6 +200,15 @@ impl<E: Env + 'static> UpdateWithCtx<E> for Player {
                 let watched_effects =
                     watched_update(&mut self.watched, &self.meta_item, &self.library_item);
 
+                let skip_gaps_effects = skip_gaps_update::<E>(
```
elpiel: This call doesn't make much sense to be honest as we depend on both the `VideoParams` and `SeriesInfo` being set/updated which for `VideoParams` only happens in a separate action.

TheBeastLT: Then probably it makes more sense to just set it to None?

src/models/player.rs
```diff
@@ -114,6 +115,8 @@ pub struct Player {
     pub paused: Option<bool>,
     #[serde(skip_serializing)]
     pub seek_history: Vec<SeekLog>,
+    #[serde(skip_serializing)]
+    pub skip_gaps: Option<(SkipGapsRequest, Loadable<SkipGapsResponse, CtxError>)>,
```
TheBeastLT, line 117: Maybe instead of exposing the raw api response, we should have some explicit structure for it and core would just make a list of skip chapters like `SkipChapter { start, end, type/name = Enum { INTRO, OUTRO } }` 
As now each client will somehow need to decide which values to use from the response, where probably core has all the data needed to decide consistently across clients which gaps need to be returned to the client.

elpiel: This will be exposed in the subsequent PR. This currently only fetches the skip gaps from the API.

elpiel: That's also the reason why it is skipped when serialising

src/models/player.rs
elpiel: Good point. It should better be set to none.

PR #582, opened 2023-12-10
feat(library_item): implement watched function

tymmesyde:
Related https://github.com/Stremio/stremio-tasks/issues/39

### DIFF
--- a/src/types/library/library_item.rs
+++ b/src/types/library/library_item.rs
@@ -63,6 +63,7 @@ impl LibraryItem {
         }
     }
 
+    /// Returns whether the item has been watched
     #[inline]
     pub fn watched(&self) -> bool {
         self.state.times_watched > 0 || self.state.flagged_watched == 1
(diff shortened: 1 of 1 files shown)


src/types/library/library_item.rs
```diff
@@ -63,6 +63,11 @@ impl LibraryItem {
         }
     }
 
+    #[inline]
+    pub fn watched(&self) -> bool {
```
elpiel, line 71: - [x] please document the fn in order for the behavior to be visible in https://stremio.github.io/stremio-core/stremio_core/

elpiel (approved):

src/types/library/library_item.rs
```diff
@@ -63,6 +63,12 @@ impl LibraryItem {
         }
     }
 
+    /// Returns whether the item has been watched
```
elpiel:
```suggestion
    /// Returns whether the item has been watched when either of the state fields are:
    /// - `times_watched > 0`
    /// or
    /// - `flagged_watched == 1` (true)
```

elpiel (approved):

PR #587, opened 2023-12-13
feat: implement dismissed events bucket

tymmesyde:


src/models/ctx/update_events.rs
```diff
@@ -1,17 +1,28 @@
...
 
-pub fn update_events<E: Env + 'static>(events: &mut Events, msg: &Msg) -> Effects {
+pub fn update_events<E: Env + 'static>(
+    events: &mut Events,
+    dismissed_events: &mut DismissedEventsBucket,
+    msg: &Msg,
+) -> Effects {
     match msg {
+        Msg::Action(Action::Ctx(ActionCtx::Logout)) | Msg::Internal(Internal::Logout) => {
+            let next_dismissed_events = DismissedEventsBucket::default();
```
elpiel, line 22: that would be a bug. You need to provide the profile UID.

src/types/events/dimissed_events_bucket.rs
```diff
@@ -0,0 +1,21 @@
...
+
+use chrono::{DateTime, Local};
+use serde::{Deserialize, Serialize};
+
+use crate::types::profile::UID;
+
+#[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct DismissedEventsBucket {
+    pub uid: UID,
+    pub items: HashMap<String, DateTime<Local>>,
```
elpiel, line 13:
```suggestion
    /// the key is the event (modal or notification) that has been dismissed
    /// while the value is the time at which it was dismissed.
    pub items: HashMap<String, DateTime<Local>>,
```

src/runtime/msg/action.rs
```diff
@@ -60,6 +60,8 @@ pub enum ActionCtx {
     PullNotifications,
     /// Make request to api to get events modal and notification
     GetEvents,
+    /// Dismiss an event
```
elpiel:
```suggestion
    /// Dismiss an event by id, either a Modal or Notification
```

src/models/ctx/update_events.rs
```diff
@@ -21,17 +32,72 @@ pub fn update_events<E: Env + 'static>(events: &mut Events, msg: &Msg) -> Effect
...
+                    eq_update(
+                        &mut events.notification,
+                        Loadable::Ready(response.to_owned()),
+                    )
+                }
+                _ => eq_update(
+                    &mut events.notification,
+                    Loadable::Ready(response.to_owned()),
+                ),
+            },
```
elpiel: - [ ] case that is not handled: Event that's dismissed should not result to `Loadable::Ready`

src/models/ctx/update_events.rs
```diff
@@ -21,17 +32,72 @@ pub fn update_events<E: Env + 'static>(events: &mut Events, msg: &Msg) -> Effect
...
+                .join(events_modal_effects)
+                .join(events_notification_effects)
+        }
         Msg::Internal(Internal::GetModalResult(_, result)) => match result {
-            Ok(response) => eq_update(&mut events.modal, Loadable::Ready(response.to_owned())),
+            Ok(response) => match response {
+                Some(GetModalResponse { id, .. }) if !dismissed_events.items.contains_key(id) => {
+                    eq_update(&mut events.modal, Loadable::Ready(response.to_owned()))
+                }
+                _ => eq_update(&mut events.modal, Loadable::Ready(response.to_owned())),
```
elpiel: Same for this match. Both match arms are equal and we need to handle the case where we have dismissed the Modal
- [ ] fix match

elpiel (commented):
How often we will call the `GetEvents`? We might need to add a timeout for making the requests e.g. every 6 or 12 hours at most.

src/models/ctx/update_events.rs
tymmesyde: The request was successful so it's Ready, not loading, not an error
If it was dismissed, it returns None

src/models/ctx/update_events.rs
tymmesyde: Check the test, if the modal was dismissed, `events.modal` will be Loadable:Ready(None)

tymmesyde: I can put explicitly Loadable::Ready(None) if you prefer

tymmesyde: No you are absolutely right, will fix that

elpiel (approved):

unclekingpin (commented):
looks great

PR #602, opened 2023-12-29
Feat/Player

elpiel:
- [x] Remove the commented JS code which is used for reference for our implementation


src/types/api/response.rs
```diff
@@ -114,3 +117,114 @@ impl LibraryItemsResponse {
...
+#[serde(rename_all = "camelCase")]
+pub struct SeekEvent {
+    pub records: u64,
+    #[serde(rename = "seekFrom")]
+    pub from: u64,
+    #[serde(rename = "seekTo")]
+    pub to: u64,
+}
+
+#[cfg(test)]
```
tymmesyde, line 173: Tests should be moved to `unit_tests`

src/models/player.rs
```diff
@@ -1063,6 +1131,227 @@ fn push_seek_to_api<E: Env + 'static>(seek_log_req: SeekLogRequest) -> Effect {
...
+            Some(video_params),
+            Some(series_info),
+            Some(library_item),
+        ) => {
+            match (
+                &selected.stream.source,
+                selected.stream.name.as_ref(),
+                video_params.hash.clone(),
+            ) {
+                (StreamSource::Torrent { .. }, Some(stream_name), Some(opensubtitles_hash)) => {
```
tymmesyde, line 1262: We should be able to handle `StreamSource::Url` (but not magnets) too

src/models/player.rs
```diff
@@ -1063,6 +1131,227 @@ fn push_seek_to_api<E: Env + 'static>(seek_log_req: SeekLogRequest) -> Effect {
...
+        video_params,
+        series_info,
+        library_item,
+        skip_gaps,
+    );
+
+    let intro_outro_effects = match (skip_gaps, library_item) {
+        (Some((_, Loadable::Ready(response))), Some(library_item)) => {
+            let outro_time = {
+                // durations = Object.keys(intro.gaps || {}).filter(function(d) {
```
tymmesyde: Those comments containing js code should probably be removed

tymmesyde:
We also need a boolean setting to enable/disable this

src/models/player.rs
elpiel: yes, this is here for context on the logic that comes from Stremio.
Will be removed before merging and I've added task for it in the PR.

src/types/api/response.rs
elpiel: We've discussed this with Niki. Adding unit tests in the module is how rust projects test. tests which are integration etc. should be put in `/tests` so we just keep the `unit_tests` for now but this approach here is more idiomatic

src/models/player.rs
elpiel: Can you elaborate on how?
Maybe this can be subsequent PR as this PR is only focusing on adding basic support taken from the Desktop app and there we only handle Magnet links.

elpiel:
> We also need a boolean setting to enable/disable this

As we discussed on slack, this is not mandatory for the time being and it's only usable if the user is premium.

The pop-up for the next episode is different or could be the same and it's not automatically skipping the outro.

Will not implement the setting in this PR.

src/models/player.rs
elpiel: I've opened #610 for this as it's not supported by desktop and needs to be tested.

tymmesyde (approved):

PR #610, opened 2024-01-08
feat: Player support of stream urls

elpiel:
Adds `StreamingSource::URL` as supported in Player


tymmesyde (changes requested):
It makes the core panick with Url streams

elpiel:
@tymmesyde I was not able to reproduce this panicking behaviour.
I get many many streams from Premiumize which just don't work because the probe endpoint and the other server-related calls fail but for the ones that it works, I can get the SeekLog be send:

image

tymmesyde (approved):

PR #613, opened 2024-01-10
Only update network_info and playback_devices on request error

tymmesyde:
Was causing the server status to be offline if one of those requests failed


elpiel (approved):
As far as I understand your comment this will not affect any other use case of the server?!

I wonder why we cleared the `torrent` field as well in these cases

tymmesyde:
> As far as I understand your comment this will not affect any other use case of the server?!
> 
> I wonder why we cleared the `torrent` field as well in these cases

Must have copy pasted this logic without thinking it about it too much
But it should not interfere with other logics

PR #621, opened 2024-01-18
fix: Addon collection and Library fetch on user login raise flags

elpiel:
## API call to retrieve Addons Collection failed on login:

image

## API call to retrieve User Library failed on login:

image

## Both requests failing on login:

image


src/runtime/msg/internal.rs
```diff
@@ -23,7 +23,19 @@ pub type CtxStorageResponse = (
     Option<LibraryBucket>,
 );
 
-pub type AuthResponse = (Auth, Vec<Descriptor>, Vec<LibraryItem>);
+#[derive(Debug)]
+pub struct CtxAuthResponse {
+    pub auth: Auth,
+    pub addons: Vec<Descriptor>,
```
unclekingpin: would be better to replace the flags with 
```pub addons: Result<Vec<>, Error>```

the same is valid for library_items too

~Enum should be called CtxAuthResult. It is used in action called Internal::CtxAuthResult~

File: src/models/ctx/ctx.rs [hunk 2]
```diff
@@ -229,66 +264,79 @@ fn authenticate<E: Env + 'static>(auth_request: &AuthRequest) -> Effect {
...
+                    })
+            };
+
+            let (addon_collection_result, datastore_library_result) =
+                future::join(addon_collection_fut, datastore_library_fut).await;
+
+            // lock if the result from fetching the addons has failed
+            let addons_locked = addon_collection_result.is_err();
+            // set the flag to true if fetching of the library has failed
+            let library_missing = datastore_library_result.is_err();
```
unclekingpin: i dont like it that errors are swallowed here
if you change the CtxAuthResponse struct as i suggested this could be handled like

  future::try_join(addon_collection_fut, datastore_library_fut)
                    .then(move |(addons_result, library_items_result)| (auth, addons_result, library_items_result))
                    
then fallback to Official_addons somwhere else, not inside the authenticate function

src/types/profile/profile.rs
```diff
@@ -15,10 +15,20 @@ pub type UID = Option<String>;
...
     #[serde_as(deserialize_as = "UniqueVec<Vec<_>, DescriptorUniqueVecAdapter>")]
     pub addons: Vec<Descriptor>,
+    /// This locking flag is raised when the API addon fetch request has failed
+    /// in order to avoid overwriting the user's addons in the API
+    /// if they install a new addon locally when we have defaulted to the official ones
+    #[serde(default)]
+    pub addons_locked: bool,
+    /// This missing library flag is raised when the API Library Collection fetch request on login
+    /// has failed to indicate why the user has an empty Library
+    #[serde(default)]
```
unclekingpin: Do we need migration if we have serde default?

elpiel: I preferred it for consistency.

src/models/ctx/ctx.rs
elpiel: Resolved

src/runtime/msg/internal.rs
```diff
@@ -23,7 +23,12 @@ pub type CtxStorageResponse = (
     Option<LibraryBucket>,
 );
 
-pub type AuthResponse = (Auth, Vec<Descriptor>, Vec<LibraryItem>);
+#[derive(Debug)]
+pub struct CtxAuthResponse {
+    pub auth: Auth,
+    pub addons_result: Result<Vec<Descriptor>, CtxError>,
```
unclekingpin, line 29: those `_result` suffixes are fine but only if we use them everywhere, on every Result and on every Option, for consistency.
might be better to get rid of em

src/models/ctx/ctx.rs
```diff
@@ -229,66 +271,80 @@ fn authenticate<E: Env + 'static>(auth_request: &AuthRequest) -> Effect {
     let auth_api = APIRequest::Auth(auth_request.clone());
 
     EffectFuture::Concurrent(
-        E::flush_analytics()
```
unclekingpin, line 232: I cannot check whats actually changed in the authenticate function. It seems ok, but ideally it shouldve been only the future join at the bottom.
```
    future::join(addon_collection_fut, datastore_library_fut).then(
        move |(addons, library_items)| {
            future::ok(CtxAuthResponse {
                auth,
                addons,
                library_items,
            })
        },
    )
```

unclekingpin (commented):
Migration is unnessesary but it wont cause any issues.
LGTM

src/models/ctx/update_profile.rs
```diff
@@ -119,6 +121,10 @@ pub fn update_profile<E: Env + 'static>(
             Effects::msg(Msg::Internal(Internal::UninstallAddon(addon.to_owned()))).unchanged()
         }
         Msg::Action(Action::Ctx(ActionCtx::UpgradeAddon(addon))) => {
+            if profile.addons_locked {
+                return addon_install_error_effects(addon, OtherError::UserAddonsAreLocked);
```
tymmesyde:
```suggestion
                return addon_upgrade_error_effects(addon, OtherError::UserAddonsAreLocked);
```

src/models/ctx/update_profile.rs
```diff
@@ -153,6 +159,10 @@ pub fn update_profile<E: Env + 'static>(
             .join(Effects::msg(Msg::Internal(Internal::ProfileChanged)))
         }
         Msg::Internal(Internal::UninstallAddon(addon)) => {
+            if profile.addons_locked {
+                return addon_install_error_effects(addon, OtherError::UserAddonsAreLocked);
```
tymmesyde:
```suggestion
                return addon_uninstall_error_effects(addon, OtherError::UserAddonsAreLocked);
```

tymmesyde (approved):

PR #626, opened 2024-01-22
refactor: use Option for audio_language and subtitles_language

tymmesyde:
Allow to set to null/to disable audio and subtitles language


elpiel (approved):
Don't we always set some defaults on purpose?
Looks good to me otherwise.

PR #634, opened 2024-01-30
Fix auth on addons or library failed and lock addons if failed

tymmesyde:


src/types/profile/profile.rs
```diff
@@ -19,6 +19,8 @@ pub struct Profile {
     pub auth: Option<Auth>,
     #[serde_as(deserialize_as = "UniqueVec<Vec<_>, DescriptorUniqueVecAdapter>")]
     pub addons: Vec<Descriptor>,
+    #[serde(skip)]
```
unclekingpin, line 22: Why serde(skip)? Addons will be automatically unlocked on the next page refresh.

src/models/ctx/update_profile.rs
```diff
@@ -311,13 +322,16 @@ pub fn update_profile<E: Env + 'static>(
...
             }
-            Err(error) => Effects::msg(Msg::Event(Event::Error {
-                error: error.to_owned(),
-                source: Box::new(Event::AddonsPulledFromAPI {
-                    transport_urls: Default::default(),
-                }),
-            }))
-            .unchanged(),
+            Err(error) => {
+                profile.addons_locked = true;
```
unclekingpin, line 326: that will cause addons locking on any failed request for pulling addons.
we do pull addons in stremio-web on window focus.
Is that expected behavior?

kKaskak:
outdated

PR #636, opened 2024-01-31
feat: add subtitles_opacity to profile settings

tymmesyde:
Related https://github.com/Stremio/stremio-video/pull/57

### DIFF
--- a/src/runtime/env.rs
+++ b/src/runtime/env.rs
@@ -658,18 +658,6 @@ mod test {
         );
     }
 
-    fn assert_storage_profile(profile: Value) {
-        let storage = STORAGE.read().expect("Should lock");
-
-        assert_eq!(
-            &profile.to_string(),
-            storage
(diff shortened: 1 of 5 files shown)


src/runtime/env.rs
```diff
@@ -628,6 +658,18 @@ mod test {
...
+        let storage = STORAGE.read().expect("Should lock");
+
+        assert_eq!(
+            &profile.to_string(),
+            storage
+                .get(PROFILE_STORAGE_KEY)
+                .expect("Should have the schema set"),
+            "Profile should match"
+        );
+    }
```
elpiel: sadly, this kind of helper functions will obscure the test and location of the error if they are used in more than 1 place in the test or multiple tests. Consider either another solution or copy-past for now

elpiel (approved):
Only just need to pick otherwise the PR looks good.

elpiel (approved):

PR #642, opened 2024-02-12
Feat: Added the extra filter option for Watched

ensapra:
After months and years of waiting for input regarding this feature request, I decided to try to implement it myself. 
# What was done?
Added a new enum called watched, that contains three options
- NotWatched
- Watched
- Any
It works exactly as the sorting but instead, it filters taking into account the watched parameter.

## Chains
https://github.com/Stremio/stremio-web/pull/583
https://github.com/Stremio/stremio-core-web/pull/97

# TODO
- [ ] Improve detection of watched/Not watched to take into account new episodes on shows. Currently for episodes it doesn't make sense that after watching one episode it is tagged as watched.
- [ ] Add translations

# Issues
https://github.com/Stremio/stremio-web/issues/582
https://github.com/Stremio/stremio-features/issues/74
https://github.com/Stremio/stremio-features/issues/552
https://github.com/Stremio/stremio-shell/issues/388

### DIFF
--- a/src/models/library_with_filters.rs
+++ b/src/models/library_with_filters.rs
@@ -63,7 +63,7 @@ pub enum Sort {
 #[derive(Derivative, Clone, PartialEq, Eq, EnumIter, Serialize, Deserialize, Debug)]
 #[derivative(Default)]
 #[serde(rename_all = "lowercase")]
-pub enum StateFilter {
+pub enum Watched {
     #[derivative(Default)]
     NotWatched,
     Watched,
@@ -76,7 +76,7 @@ pub struct LibraryRequest {
     #[serde(default)]
     pub sort: Sort,
     #[serde(default)]
-    pub stateFilter: StateFilter,
+    pub watched: Watched,
     #[serde(default)]
     pub page: LibraryRequestPage,
 }
@@ -110,8 +110,8 @@ pub struct SelectableSort {
 }
 
 #[derive(Clone, PartialEq, Eq, Serialize, Debug)]
-pub struct SelectableStateFilter {
-    pub stateFilter: StateFilter,
+pub struct SelectableWatched {
+    pub watched: Watched,
     pub selected: bool,
     pub request: LibraryRequest,
 }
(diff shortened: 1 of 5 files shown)


src/models/library_with_filters.rs
```diff
@@ -60,12 +60,24 @@ pub enum Sort {
     TimesWatched,
```
tymmesyde, line 60: There is already an enum for the filters called `Sort`, you should add `NotWatched` here

src/models/library_with_filters.rs
```diff
@@ -315,6 +368,11 @@ fn catalog_update<F: LibraryFilter>(
...
                 Some(r#type) => library_item.r#type == *r#type,
                 None => true,
+            })          
+            .filter(|library_item| match &selected.request.watched {
+                Watched::NotWatched => !library_item.watched(),
+                Watched::Watched => library_item.watched(),
+                Watched::Any => true,
             })
             .sorted_by(|a, b| match &selected.request.sort {
                 Sort::LastWatched => b.state.last_watched.cmp(&a.state.last_watched),
```
tymmesyde, line 378: And add the filter logic of  `Sort:Watched` here

ensapra: I'm not interested in sorting, I'm interested in a filter to reduce the amount of items shown in the library. By making the default not watched, it's really easy to keep it clean and only see new things on it.

ensapra: But I'm not sure if you are suggesting to add the options in sort but still be able to do a filter?

ensapra: Keeping it separated also helps to be able to sort and filter at the same time, so we don't lose features

tymmesyde: Ok i misunderstood then, can you rename `Watched` to `Filter` instead and every other instances ?

ensapra: Sounds good

ensapra: I updated the repos with the corresponding change

elpiel:
Closing in favour of #647 
Thanks a lot @ensapra , it's time to get this feature to web :yum:

PR #643, opened 2024-02-27
Implement LoadNextPage for LibraryWithFilters

tymmesyde:
Update the pagination logic to just append more items with LoadNextPage instead of returning slices
Related https://github.com/Stremio/stremio-core-web/pull/99


elpiel (approved):
Tested with CATALOG_PAGE_SIZE of 14 and works using the other 2 PRs in web and core-web.

PR #645, opened 2024-03-01
Feat/improve serde error handling

elpiel:
- Introduce additional details in Addons' ResourceResponse when deserialization fails
- Introduce `serde_path_to_error` which, if addon returns a badly formatted JSON, will result in a clear error where the mistake is located
**NB:** On items inside each of the different response types we still skip any item which is not deserializable (e.g. missing a mandatory field)

- Refactor APIResult to use tagged enum variant (related to API change https://github.com/Stremio/stremio-api/pull/184, `error` is no longer set if it's `null`)
Given a bad response of `{"unknown_variant": {"test": 1}}`
When `APIResult` does not match any variant it returns a vary vague error:
`data did not match any variant of untagged enum APIResult`

with this change we now return a much cleaner and well understood error:
``unknown variant `unknown_variant`, expected `error` or `result` at line 1 column 18``


Ivshti (approved):

src/types/addon/response copy.rs
tymmesyde, line 1: Should be removed or merged with `addon/response.rs`

elpiel: It was accidentally committed. Will remove it

Ivshti (approved):

tymmesyde (approved):

PR #647, opened 2024-03-07
Feat/sort not watched and watched

elpiel:
@ensapra this is continuation from your PR #642 that uses the existing `Sort` structure instead of creating a Filter.
Thank you very much for your contribution!
We will be discussing the UI for web internally as well in hopes to improve the UX for using these sort options in the UI.

### DIFF
--- a/src/models/library_with_filters.rs
+++ b/src/models/library_with_filters.rs
@@ -358,8 +358,7 @@ fn catalog_update<F: LibraryFilter>(
                 None => true,
             })
             .sorted_by(|a, b| selected.request.sort.sort_items(a, b))
-            .skip((selected.request.page.get() - 1) * CATALOG_PAGE_SIZE)
-            .take(CATALOG_PAGE_SIZE)
+            .take(selected.request.page.get() * CATALOG_PAGE_SIZE)
             .cloned()
             .collect(),
         _ => vec![],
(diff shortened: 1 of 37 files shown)


src/models/library_with_filters.rs
```diff
@@ -57,6 +57,62 @@ pub enum Sort {
...
+                    match (&b.ctime, &a.ctime) {
+                        (Some(b_ctime), Some(a_ctime)) => b_ctime.cmp(a_ctime),
+                        // if one or both are missing, we just leave the ordering by default
+                        _ => Ordering::Equal,
+                    }
+                }
+            }
+            ordering => ordering,
+        }
+    }
```
tymmesyde: Could be simplified to this
```suggestion
            Sort::Watched => b
                .watched()
                .cmp(&a.watched())
                .then(b.state.last_watched.cmp(&a.state.last_watched))
                .then(b.ctime.cmp(&a.ctime)),
            Sort::NotWatched => a
                .watched()
                .cmp(&b.watched())
                .then(b.state.last_watched.cmp(&a.state.last_watched))
                .then(b.ctime.cmp(&a.ctime)),
            Sort::Name => a.name.to_lowercase().cmp(&b.name.to_lowercase()),
        }
    }
```

tymmesyde (approved):

src/models/library_with_filters.rs
elpiel: Thank you very much for the suggestion. It works and I've added an additional item to test a watched movie (with last_watched but flagged_watched == 0) and it's ordering.

ensapra:
Omg, I'm so happy right now. Thank you for listening, I can't wait to see this feature available on the web!

PR #650, opened 2024-03-11
fix: mark as watched behavior for meta_details

tymmesyde:
Closes https://github.com/Stremio/stremio-tasks/issues/203

- Don't forward the event as `LibraryItemMarkAsWatched`
- Fix the condition for the watched function of library_item
- Use `flagged_watched` instead of updating `times_watched`

### DIFF
--- a/src/models/meta_details.rs
+++ b/src/models/meta_details.rs
@@ -110,11 +110,11 @@ impl<E: Env + 'static> UpdateWithCtx<E> for MetaDetails {
             Msg::Action(Action::MetaDetails(ActionMetaDetails::MarkAsWatched(is_watched))) => {
                 match &self.library_item {
                     Some(library_item) => {
-                        Effects::msg(Msg::Internal(Internal::LibraryItemMarkAsWatched {
-                            id: library_item.id.clone(),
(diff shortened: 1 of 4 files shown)


elpiel (approved):
LGTM!

elpiel:
> i accept tim's proposal to not increment times watched.. but maybe we should just not increment it over 1? (yet still change it from 0 to 1 the first time u mark as watched)
From @jaruba 

I'm for the chagne, @tymmesyde if you agree, let's just make the change and merge this PR.

unclekingpin (changes requested):
I have issues with this PR.
flagged_watched is not the best name for this flag, but its supposed to be used only in the player, when the user reach some threshold so that we dont do unnesesary operation in the watched bitfield.

How can i reproduce this https://github.com/Stremio/stremio-tasks/issues/203
The only way i could think of is if the times_watched counter already had u32::MAX and we try to do library_item.state.times_watched += 1; instead of library_item.state.times_watched.saturating_add(1)

elpiel:
#674 merged instead, this PR is no longer needed

PR #656, opened 2024-03-15
feat(library_with_filters): add NameReverse sort filter

tymmesyde:
Related to https://github.com/Stremio/stremio-tasks/issues/188


elpiel (approved):
I wanted to make this NameDescending (or NameDesc) but since desktop also uses the naming `reverse` let's keep it this way

PR #663, opened 2024-03-21
feat: handle device-info route for streaming server

tymmesyde:


src/types/streaming_server/settings.rs
```diff
@@ -7,6 +7,7 @@ pub struct Settings {
     pub cache_root: String,
     pub server_version: String,
     pub remote_https: Option<String>,
+    pub transcode_profile: Option<String>,
```
elpiel, line 15: Please document what this value is used for and some examples of what it can include

src/types/streaming_server/device_info.rs
```diff
@@ -0,0 +1,7 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, PartialEq, Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct DeviceInfo {
```
elpiel, line 7: Would be nice to also include documentation.
I suppose this is information of the device running the  server?!

src/types/streaming_server/settings.rs
```diff
@@ -7,6 +7,7 @@ pub struct Settings {
     pub cache_root: String,
     pub server_version: String,
     pub remote_https: Option<String>,
```
elpiel, line 11: Would be nice to include some documentation for remote_https as well, e.g. what this field means and 1-2 examples

elpiel (commented):
LGTM in general, nothing much to add than adding docs.

src/types/streaming_server/settings.rs
tymmesyde: Can we document this server side then ? Because it's related to the streaming server and if we start documenting this field we should document every one of them

src/types/streaming_server/device_info.rs
```diff
@@ -0,0 +1,7 @@
...
+    pub available_hardware_accelerations: Vec<String>,
```
elpiel, line 10: - [x] FIXME: the server can actually return `false` when there are no available HW accelerations

src/types/streaming_server/settings.rs
elpiel: usually - yes, we should. At least we need to know what it does and what is it usescase.
let's leave it without a documentation for this time.

PR #664, opened 2024-03-22
Fix next video stream

tymmesyde:


elpiel (approved):

src/types/resource/stream.rs
```diff
@@ -228,14 +228,8 @@ impl Stream {
...
-    pub fn is_binge_match(&self, other_stream: &Stream) -> bool {
-        match (
-            &self.behavior_hints.binge_group,
-            &other_stream.behavior_hints.binge_group,
-        ) {
-            (Some(a), Some(b)) => a == b,
-            _ => false,
-        }
+    pub fn is_binge_match(&self, stream: &Stream) -> bool {
+        self.behavior_hints.binge_group == stream.behavior_hints.binge_group
```
TheBeastLT, line 232: This is incorrect as it will return true when both `binge_group` are empty, and that should not be the case.

tymmesyde: Right this should not have been included in this PR

PR #675, opened 2024-04-22
fix(Player): LibraryItem marked as watched on TimeChanged

elpiel:


src/models/player.rs
```diff
@@ -432,8 +432,9 @@ impl<E: Env + 'static> UpdateWithCtx<E> for Player {
                     };
                     library_item.state.time_offset = time.to_owned();
                     library_item.state.duration = duration.to_owned();
+
                     if library_item.state.flagged_watched == 0
-                        && library_item.state.time_watched as f64
+                        && library_item.state.time_offset as f64
```
TheBeastLT, line 437: I don't think this has changed, or causes bugs, since it was always like this - https://github.com/Stremio/stremio-models/blob/a0a5055f1392361dfe32530df030c29c95cb6f65/libraryItem.js#L338
With this change it would mean that you could seek to end of the video and it would mark it as watched, where actually you didn't watch it.

elpiel: yes, but you are missing the case where people have seeked more than 20% in an episode (think of skipping end scenes and intro) which will make the time_watched (current played time) less than 80% and will not mark the episode as watched.

The user doesn't mention how much of the episode they are continue watching or how much of it was skipped but the episodes of Modern Family are 20 minutes so it's easy to skip it.

Another issue with this approach is that if you seek backward:

```
let time_watched =
                            1000.min(time.saturating_sub(library_item.state.time_offset));
```

It won't take into account going backward and if you've, for example, fell a sleep and watched to some point, it is possible that when you go back and rewind it will mark it earlier as watched.

TheBeastLT: The coef is at 70% to take all the possible skipping into account, which I think is reasonable. Also all of this logic was defined a long long time ago (used in old stremio-models as well) and worked well, so I really doubt it's related in any way to the new issues.

elpiel: There is one other change that is related to marking an item as watched - `item_state_update` it's coef. is 90% I honestly don't see anything else related to this issue. Literally nothing...

There are 2 checks - one for flagged_watched and one for the watched coef.

elpiel:
As discussed internally, this issue was related the Theater's throttling function which cause unexpected behavior as sending TimeChanged every ~3000 ms caused core to believe that the user is seeking and my default was using `1000.min(time_difference)` where `time_difference > 3000`, hence the time_watched was wrong and episode wasn't marked as watched.

PR #680, opened 2024-05-02
chore: Default fields for resources on `null`

elpiel:
- MetaItemPreviewLegacy
- MetaItem
- Video
- addon ResourceResponse - null as empty vec
- Stream
- impl SerializeAs for SortedVec


src/types/addon/response.rs
```diff
@@ -86,28 +86,38 @@ pub struct ResourceResponseCache {
 #[serde(untagged)]
 #[serde_as]
 pub enum ResourceResponse {
+    /// [`MetaItemPreview`] response without the `videos` key of an item
```
tymmesyde, line 89: Shouldn't be here ?

elpiel: No, it should be.
This is the documentation for the ResourceResponse and if an addon developer wants to see information about the struct this is a good place to have it.

We should also add it to the MetaItemPreview.

Keep in mind these are only the titles of sorts, to the variants.
This is no means a good documentation by itself but it is a start.

tymmesyde: What i meant is that this line should be here instead: https://github.com/Stremio/stremio-core/pull/680/files/4e81944e58dc9ccc6832ab48e77e2587f2dfacd4#diff-765e27563ea3efb23deeb517538cda76bfc1468b413d2f404c91323f707979d1R104
This enum is `ResourceResponse`

elpiel: Imo it should be at both.
Most of the documentation should be there but there should be also at least a sentence in ResourceResponse as well

elpiel: I'm keen to changing the comment to mention which endpoint this response is meant for.

tymmesyde (approved):

PR #684, opened 2024-05-09
Feat/Convert stream structs prepartion

elpiel:
A big leap forward on the effort to make all StreamSource state transitions valid.
This is the first step before encoding the state transitions to the type system.


src/models/streaming_server.rs
```diff
@@ -732,3 +710,253 @@ fn update_remote_url<E: Env + 'static>(
...
+    let filename_fut = async move {
+        match (
+            transport_url.to_owned(),
+            stream.source,
+            stream.behavior_hints.filename,
+        ) {
+            (_, _, Some(filename)) => Ok(Some(filename)),
+            (
+                Some(transport_url),
+                ConvertedStreamSource::Torrent {
```
TheBeastLT: For URL source we should take the last path segment.

src/types/streaming_server/request.rs
```diff
@@ -0,0 +1,323 @@
...
+    pub info_hash: InfoHash,
+    pub file_idx: u64,
+}
+
+impl From<CreateTorrentRequest> for Request<CreatedTorrent> {
+    fn from(val: CreateTorrentRequest) -> Self {
+        let url = {
+            let mut uri = val
+                .server_url
+                .join(&format!("{}/{}", val.info_hash, val.file_idx))
```
TheBeastLT: This is the url for a stateless GET request, for a POST request this is the endpoint AFAIK - https://github.com/Stremio/enginefs/blob/dbac17e737cf3cc25e0649f3bfee15cd64962784/enginefs.js#L354

src/types/streaming_server/request.rs
```diff
@@ -0,0 +1,323 @@
...
+impl From<CreateMagnetRequest> for Request<CreateMagnetBody> {
+    fn from(val: CreateMagnetRequest) -> Self {
+        let info_hash = val.info_hash;
+
+        let body = CreateMagnetBody {
+            torrent: CreateMagnetTorrent {
+                info_hash: val.info_hash.to_owned(),
+            },
+            peer_search: if !val.announce.is_empty() {
+                Some(PeerSearch {
```
TheBeastLT: Why not use `PeerSearch::new` here?

src/types/streams/converted_source.rs
```diff
@@ -0,0 +1,26 @@
...
+
+/// Trait which defines the StreamSource state data structures in Core.
+pub trait StreamSourceTrait: sealed::Sealed {}
+/// only we should be able to define which data structures are StreamSource states!
+mod sealed {
+    pub trait Sealed {}
+}
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum ConvertedStreamSource {
```
TheBeastLT, line 13: I'm not sure about having this struct, seems like a redundant one duplicating properties. Since The conversion end result is an url maybe just having a `ResourceLoadable<Url>` in player model would be enough, or have some specific deeplinks method which would be async and would add the url to the `deeplinks.externalLink.streaming` or smth?

src/types/streaming_server/request.rs
elpiel: the `new` function was created after the fact.
Updated!

src/types/streams/converted_source.rs
elpiel: We talked about this in Slack, tldr is that we want to have explicit state conversions, this is not used but I'll make sure to slowly transition core to this state-transition logic to make it explicit.

src/types/streaming_server/request.rs
elpiel: I've removed this request as it was not being used at the time being.

TheBeastLT (approved):

PR #685, opened 2024-05-10
do not double encode path segments in proxied url with headers

TheBeastLT:


src/types/resource/stream.rs
```diff
@@ -212,12 +212,9 @@ impl Stream {
...
                         );
-                        streaming_url
-                            .path_segments_mut()
-                            .ok()?
-                            .push("proxy")
-                            .push(proxy_query.finish().as_str())
-                            .push(&url.path()[1..]);
+                        streaming_url.set_path(
+                            &["proxy", proxy_query.finish().as_str(), &url.path()[1..]].join("/"),
+                        );
```
elpiel:
```suggestion
                        streaming_url.set_path(&format!(
                            "proxy/{query}/{url_path}",
                            query = proxy_query.finish().as_str(),
                            url_path = &url.path().strip_prefix('/').unwrap_or(&url.path()),
                        ));
```

PR #697, opened 2024-06-18
Calendar model

tymmesyde:


src/models/calendar.rs
```diff
@@ -0,0 +1,338 @@
...
+    pub year: Year,
+}
+
+impl Date {
+    pub fn new(year: Year, month: Month, day: Option<Day>) -> Self {
+        Self { day, month, year }
+    }
+}
+
+impl From<NaiveDate> for Date {
```
elpiel: This is confusing. NaiveDate also has a day. Why is it different that the DateTime? 

- [x] Please handle  Day vs Month values differently

TRtomasz:
imo no point of making a pr that does not include all needed dependencies to run. It is untestable. I this case it  should include in `package.json` core-web version that is still in PR without waiting for it to be merged.  Preferred solution would be to make this pr only after core and core-web changes are merged so we could avoid testing something without a reason in case it is decided that core/core-web changes should be rewritten or changed in a bigger part.

elpiel:
There are a few things in the PR that I would like to change. Especially with handling the Date changes and date changes with offset. I've worked on some of the changes but left it as I had to work on more pressing issues.

Will get back to this.

src/models/calendar.rs
```diff
@@ -0,0 +1,338 @@
...
+    pub date: Date,
+    pub items: Vec<ContentItem>,
+}
+
+#[derive(Default, Clone, PartialEq, Eq, Serialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct MonthInfo {
+    pub today: Option<Day>,
+    pub days: u32,
+    pub first_weekday: u32,
```
elpiel, line 103: is this first_weekday indexed from 0 or from 1?
This usually can easily by extracted from methods on Date or DateTime instances, including how many days a certain month has.

src/models/calendar.rs
```diff
@@ -0,0 +1,338 @@
...
+
+    while month > 12 {
+        month -= 12;
+        year += 1;
+    }
+
+    while month < 1 {
+        month += 12;
+        year -= 1;
+    }
```
elpiel: This is not how you should do it, especially the while loops..
There are methods that do this for you.

elpiel (changes requested):

elpiel:
I will still test the stremio-web for the UI implementation and general workings of the logic but there are a few things that need refactoring from the model.
Logic looks good so far though.

src/models/calendar.rs
```diff
@@ -0,0 +1,338 @@
...
+            };
+
+            let date = NaiveDate::from_ymd_opt(*year, *month, 1).unwrap_or_default();
+            let first_of_next_month = date_from_month_offset(date, 1);
+
+            let last_day_of_month = first_of_next_month.pred_opt().unwrap_or_default();
+            let days = last_day_of_month.day();
+
+            let first_day_of_month = date.with_day(1).unwrap_or_default();
+            let first_weekday = first_day_of_month.weekday().num_days_from_monday();
```
elpiel:
```suggestion
            let first_day_of_month = NaiveDate::from_ymd_opt(*year, *month, 1).unwrap_or_default();
            let first_of_next_month = first_day_of_month + Months::new(1);
            let last_day_of_month = first_of_next_month.pred_opt().unwrap_or_default();

            let days = last_day_of_month.day();

            let first_weekday = first_day_of_month.weekday().num_days_from_monday();
```

src/models/calendar.rs
```diff
@@ -0,0 +1,338 @@
...
+}
+
+fn selectable_update(selectable: &mut Selectable, selected: &Option<Selected>) -> Effects {
+    let updated_selectable = selected
+        .as_ref()
+        .map(|Selected { month, year, .. }| {
+            let date = NaiveDate::from_ymd_opt(*year, *month, 1).unwrap_or_default();
+
+            let prev_date = date_from_month_offset(date, -1);
+            let next_date = date_from_month_offset(date, 1);
```
elpiel:
```suggestion
            let prev_date = date - Months::new(1);
            let next_date = date + Months::new(1);
```

src/models/calendar.rs
tymmesyde: From 0
It's using `.weekday().num_days_from_monday()` from NaiveDate 
https://github.com/Stremio/stremio-core/pull/697/files#diff-2e83cb7dcf42829873903acb76fb7f66c5b91ddde7ab289eaf45f2013e788048R207

elpiel (approved):

PR #699, opened 2024-06-23
proxy_streams_enabled added to streaming server settings

unclekingpin:


unclekingpin:
@elpiel merged this directly to development, please review post merge

src/types/streaming_server/settings.rs
```diff
@@ -14,6 +14,7 @@ pub struct Settings {
     #[serde_as(deserialize_as = "DefaultOnError")]
     pub transcode_profile: Option<String>,
     pub cache_size: Option<f64>,
+    pub proxy_streams_enabled: bool,
```
elpiel, line 17: If this field is camelCase and it's always boolean on the server side and there is no Null option returned (ever), then PR seems ok!
I will try to check the server code as well

elpiel: Seems like it so PR is ok.

PR #702, opened 2024-07-08
fix(core): StreamingServer Settings.remote_https uses null while stremio 4 uses empty string, with tests

DoomHEADSHOT:
fix for this: https://github.com/Stremio/stremio-tasks/issues/174


src/types/streaming_server/settings.rs
```diff
@@ -22,3 +23,112 @@ pub struct Settings {
...
+            bt_handshake_timeout: 30,
+            bt_request_timeout: 60,
+            bt_download_speed_soft_limit: 100.0,
+            bt_download_speed_hard_limit: 200.0,
+            bt_min_peers_for_stable: 5,
+        };
+
+        let json = serde_json::to_string(&settings).unwrap();
+
+        println!("JSON: {}", json);
```
elpiel: - [x] No println! in tests please.

src/types/streaming_server/settings.rs
```diff
@@ -22,3 +23,112 @@ pub struct Settings {
...
+
+        assert_eq!(
+            json,
+            r#"{"appPath":"app_path","cacheRoot":"cache_root","serverVersion":"server_version","remoteHttps":"","transcodeProfile":null,"cacheSize":null,"proxyStreamsEnabled":true,"btMaxConnections":10,"btHandshakeTimeout":30,"btRequestTimeout":60,"btDownloadSpeedSoftLimit":100.0,"btDownloadSpeedHardLimit":200.0,"btMinPeersForStable":5}"#
```
elpiel: This is error prone! Please use the serde_json::json macro to create the json instead.

src/types/streaming_server/settings.rs
```diff
@@ -22,3 +23,112 @@ pub struct Settings {
...
+    fn test_deserialize_remote_https_none() {
+        let json = r#"{"appPath":"app_path","cacheRoot":"cache_root","serverVersion":"server_version","remoteHttps":"","transcodeProfile":null,"cacheSize":null,"proxyStreamsEnabled":true,"btMaxConnections":10,"btHandshakeTimeout":30,"btRequestTimeout":60,"btDownloadSpeedSoftLimit":100.0,"btDownloadSpeedHardLimit":200.0,"btMinPeersForStable":5}"#;
+
+        let settings: Settings = serde_json::from_str(json).unwrap();
+
+        assert_eq!(settings.remote_https, None);
+    }
+
+    #[test]
+
```
elpiel: empty line

src/types/streaming_server/settings.rs
```diff
@@ -22,3 +23,112 @@ pub struct Settings {
...
+        if s.is_empty() {
+            Ok(None)
+        } else {
+            Ok(Some(s))
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
```
elpiel: Move the module inside `empty_string_as_null`
Also make sure to:
- [x] use serde_josn::json to generate **valid** json instead of relying on strings like now.
- [x] Instead of using Settings, just create your own structure in the test module, e.g.:
```rust
        #[derive(Debug, Deserialize, Serialize, PartialEq, Eq)]
        struct TestStruct {
            #[serde(with = "super")]
            field: Option<String>,
        }
```

You can still use the remote_https if you want but the behavior of the module is generic, just keep that in mind when writing the test cases, i.e. more generic name of the test of what's actually testing is a good rule of thumb.

elpiel (changes requested):
Overall LGTM, a few tweaks and clean up is needed before merging.

src/types/streaming_server/settings.rs
```diff
@@ -1,13 +1,14 @@
+use crate::types::streaming_server::serde_ext::empty_string_as_null;
 use serde::{Deserialize, Serialize};
 use serde_with::{serde_as, DefaultOnError};
```
elpiel:
```suggestion
use serde::{Deserialize, Serialize};
use serde_with::{serde_as, DefaultOnError};

use crate::types::streaming_server::serde_ext::empty_string_as_null;

```

src/types/streaming_server/serde_ext.rs
elpiel, line 1: - [x] Please add another test, as your current implementation fails on values that are `{ "field": null }`. `fn deserialize` should be fixed.

elpiel (changes requested):

src/types/streaming_server/serde_ext.rs
```diff
@@ -0,0 +1,83 @@
...
+        D: Deserializer<'de>,
+    {
+        let s: Option<String> = Deserialize::deserialize(deserializer)?;
+        match s {
+            None => Ok(None),
+            Some(s) => match s.is_empty() {
+                true => Ok(None),
+                false => Ok(Some(s)),
+            }
+        }
```
elpiel:
```suggestion
        let s = Option::<String>::deserialize(deserializer)?;
        match s {
            Some(s) if !s.is_empty() => Ok(Some(s))
            _ => Ok(None)
        }
```

elpiel:
Tested and works with the web changes made in:
https://github.com/Stremio/stremio-web/pull/663

elpiel (approved):

kKaskak (approved):

elpiel (approved):

PR #709, opened 2024-07-26
feat: ExternalPlayerLink - Add moonplayer, vlc and outplayer for visionOS

elpiel:
Adds the Moonplayer, VLC and Outplayer (for the Vision Pro) as an external player

### DIFF
--- a/src/deep_links/mod.rs
+++ b/src/deep_links/mod.rs
@@ -35,7 +35,7 @@ pub struct OpenPlayerLink {
     pub chromeos: Option<String>,
     pub roku: Option<String>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    /// Vision Pro os
+    /// VisionOS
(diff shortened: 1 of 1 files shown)


src/deep_links/mod.rs
```diff
@@ -34,6 +34,9 @@ pub struct OpenPlayerLink {
     pub webos: Option<String>,
     pub chromeos: Option<String>,
     pub roku: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    /// Vision Pro os
```
kKaskak:
```suggestion
    /// VisionOS
```

kKaskak (changes requested):

tymmesyde:
`visionos` fields are missing for VLC and Outplayer
We can probably assume that they are the same as ios

kKaskak (approved):

PR #724, opened 2024-10-03
feat: streaming-server urls bucket

kKaskak:


src/types/server_urls/server_urls_bucket.rs
```diff
@@ -0,0 +1,73 @@
...
+        let base_url: &Url = &STREAMING_SERVER_URL;
+        let mtime = Self::current_timestamp();
+        items.insert(base_url.clone(), mtime);
+
+        ServerUrlsBucket { uid, items }
+    }
+
+    fn current_timestamp() -> DateTime<Utc> {
+        chrono::Utc::now()
+    }
```
elpiel: - [x] if you need timestamp you need to use the Env impl

src/unit_tests/ctx/update_streaming_server_urls.rs
```diff
@@ -0,0 +1,111 @@
...
+use crate::types::profile::Profile;
+use crate::types::search_history::SearchHistoryBucket;
+use crate::types::server_urls::ServerUrlsBucket;
+use crate::types::streams::StreamsBucket;
+use crate::unit_tests::{TestEnv, STORAGE};
+use stremio_derive::Model;
+use url::Url;
+
+#[test]
+fn test_add_server_url() {
```
elpiel, line 17: - [x] Please add a test for the max number of items and the replacement (or not) of the oldest url

src/types/server_urls/server_urls_bucket.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// Add a new URL to the bucket.
+    pub fn add_url(&mut self, url: Url) {
+        let mtime = Self::current_timestamp();
+        self.merge_items(vec![(url, mtime)]);
```
tymmesyde: The use of this function seems unnecessary and could just be replaced by this since the key is the Url
```
self.items.insert(url, mtime);
```

src/unit_tests/ctx/update_streaming_server_urls.rs
kKaskak: Not needed since the logic will be removed and we will allow users to place new URLs without constrains

stremio-core-web/src/model/serialize_ctx.rs
```diff
@@ -30,6 +30,15 @@ mod model {
...
         pub search_history: Vec<SearchHistoryItem<'a>>,
         pub events: &'a Events,
+        pub streaming_server_urls: Vec<StreamingServerUrlItem>,
+    }
+
+    #[derive(Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct StreamingServerUrlItem {
+        pub url: String,
+        #[serde(rename = "_mtime")]
```
elpiel: - [x] this renaming is not necessary. For libraryItem it;'s renamed for historical reasons. This is not the case for this filed

stremio-core-web/src/model/serialize_ctx.rs
```diff
@@ -30,6 +30,15 @@ mod model {
         pub notifications: Notifications<'a>,
         pub search_history: Vec<SearchHistoryItem<'a>>,
         pub events: &'a Events,
+        pub streaming_server_urls: Vec<StreamingServerUrlItem>,
+    }
+
+    #[derive(Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct StreamingServerUrlItem {
+        pub url: String,
```
elpiel: - [x] No need for a String here, Url is serialized to string:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=80b575076c91b41ecd0443570bd9cedd

src/models/ctx/update_streaming_server_urls.rs
```diff
@@ -0,0 +1,64 @@
+use crate::constants::STREAMING_SERVER_URLS_STORAGE_KEY;
+use crate::runtime::msg::{Action, ActionCtx, CtxAuthResponse};
+use crate::runtime::EnvFutureExt;
+use crate::runtime::{
+    msg::{Event, Internal, Msg},
+    Effect, EffectFuture, Effects, Env,
+};
+use crate::types::server_urls::ServerUrlsBucket;
+use futures::FutureExt;
```
elpiel: From here on try to keep the imports consistent with general guides of ordering:

1. std
2.  other crates
3. use crate::{...} in one block

src/types/server_urls/server_urls_bucket.rs
```diff
@@ -0,0 +1,39 @@
...
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use url::Url;
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq)]
+pub struct ServerUrlsBucket {
+    /// User ID
+    pub uid: UID,
+    /// HashMap where key is `Url`, value is `DateTime<Utc>`
+    pub items: HashMap<Url, DateTime<Utc>>,
```
elpiel, line 12: There's one issue with the hashmap - it doesn't keep the ordering.
If you need for the items' order to be kept you can check out e.g. indexmap crate.

- [x] confirm whether or not we need to keep the order of items

elpiel (commented):
Looks good to me, just a few things to take into account.
Will test the UI now

src/types/server_urls/server_urls_bucket.rs
```diff
@@ -0,0 +1,39 @@
...
+use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use url::Url;
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq)]
+pub struct ServerUrlsBucket {
+    /// User ID
+    pub uid: UID,
+    /// HashMap where key is `Url`, value is `DateTime<Utc>`
```
elpiel:
```suggestion
    /// A map of the the server Url as a key and the modified/added datetime of that particular url.
```

src/types/server_urls/server_urls_bucket.rs
elpiel: Done by comparing the modified(or added) date time

elpiel (approved):

PR #736, opened 2024-11-12
Remote addons descriptor

tymmesyde:


src/types/addon/response.rs
```diff
@@ -118,7 +118,7 @@ pub enum ResourceResponse {
         subtitles: Vec<Subtitles>,
     },
     Addons {
-        addons: Vec<DescriptorPreview>,
+        addons: Vec<Descriptor>,
```
elpiel, line 121: Per SDK docs it seems we indeed use the addons indeed should return the full descriptor:
https://github.com/Stremio/stremio-addon-sdk/blob/master/docs/api/responses/addon_catalog.md

I will ping @jaruba about this just to confirm as he might know edge cases where addons return DescriptorPreview instead. I just hope we don't break any addons :D

PR #750, opened 2024-12-07
feat: add player MarkVideoAsWatched action

tymmesyde:


kKaskak (approved):

src/models/player.rs
```diff
@@ -1052,10 +1077,7 @@ fn library_item_update<E: Env + 'static>(
             meta_request: Some(meta_request),
             ..
         }) => {
-            let library_item = library_item
-                .as_ref()
-                .filter(|library_item| library_item.id == meta_request.path.id)
```
elpiel, line 1057: I'm not sure why this check was here in the first place?! Maybe this fn is called when you change the LIbaryItem and you can have a different metaRequest id?

src/models/player.rs
```diff
@@ -658,6 +658,31 @@ impl<E: Env + 'static> UpdateWithCtx<E> for Player {
...
+                        library_item.mark_video_as_watched::<E>(watched, video, *is_watched);
+
+                        Effects::msg(Msg::Internal(Internal::UpdateLibraryItem(library_item)))
+                            .unchanged()
+                    }
+                    _ => Effects::none().unchanged(),
+                }
+            }
+            Msg::Internal(Internal::LibraryChanged(_)) => {
+                let library_item_effects = library_item_update::<E>(
```
elpiel, line 674: - [x] ~~let's not update the LibraryItem if it hasn't changed in the library.~~

src/models/player.rs
elpiel, line 1057: - [x] Before removing this we need to know why it was here in the first place and if this case still holds.

Please elaborate

src/models/player.rs
```diff
@@ -1077,28 +1099,7 @@ fn library_item_update<E: Env + 'static>(
         }
         _ => None,
     };
-    if *library_item != next_library_item {
```
elpiel, line 1080: I suppose this part was moved above with the `LibraryChanged` internal message and that's why you've removed the logic here?

src/models/player.rs
tymmesyde: This check means if it's already the same library item id as the request id, don't get the library item from the library, but rather use the one that is already there on the model 
We don't want this because the library item was updated after being marked as watched

src/models/player.rs
tymmesyde: I do not really understand what was the purpose of this code because instead of updating the library item inside the model, it just dispatch the internal message to update it in the library
We don't want that either, we want to update the library item from the model after it was updated from the library

elpiel:
:heavy_check_mark: Tested and works with stremio-web build.

kKaskak:
Tested and works well in the web app ✅

src/models/player.rs
elpiel: If the LibaryItem hasn't changed, then eq_update will properly skip any updating an NewState to the model

src/models/player.rs
elpiel: Taking a second look after discussing the flow with Tim, I think this should be fine.
THe LibraryItem form the LIbrary will be the most up-to-date library item as we send any update to be updated from the Ctx

elpiel (approved):

src/models/player.rs
nklhtv: for the first 90 seconds of the Player lifecycle a library_item exists only in the Player, but not in ctx.library. (if you watch a movie for the first time ever)
All changes made to that temp library_item during those 90 seconds are gonna be lost. (in case youve marked video as watched)

nklhtv: also `self.push_library_item_time` should be reset on Unload i think.
Perhaps you get library item from the ctx.library every time bc you dont reset push_library_item_time, which causes the library_item to be persisted even on the first TimeChanged action.
This is a case which i think we wanted to avoid, watching less than 90 seconds of stream should not end up in your library(only as `temp` library_item)

elpiel: And   "only as temp" meaning "existing only in the player until those 90 seconds end"?

At which point it will place the library in your Library Bucket.

nklhtv: yes, the library_item exits in the player for the first 90 seconds (or until you dispatch some action that triggers UpdateLibraryItem, like markvideoaswatch etc)
but with this change it recreates library_item from meta_item every time, until its not actually persisted in the library bucket

PR #753, opened 2024-12-18
feat: add ctx delete account action

tymmesyde:
Closes https://github.com/Stremio/stremio-tasks/issues/300

Changed the handling of Logout messages, now when `CtxAction::Logout` is dispatched, it will dispatch `Internal::Logout`
https://github.com/Stremio/stremio-core/pull/753/files#diff-30ea419c3fb216cc24c692f6ccd053854151825012d9a17bf285525e276e50ffR117

This allow to pass a flag to `Internal::Logout` to tell if the session was already deleted server-side, to prevent sending a `delete_session` request, as the `delete_account` request already delete user session server-side
https://github.com/Stremio/stremio-core/pull/753/files#diff-30ea419c3fb216cc24c692f6ccd053854151825012d9a17bf285525e276e50ffR123

### DIFF
--- a/src/models/ctx/update_profile.rs
+++ b/src/models/ctx/update_profile.rs
@@ -30,8 +30,8 @@ pub fn update_profile<E: Env + 'static>(
                 Effects::none().unchanged()
             }
         }
-        Msg::Action(Action::Ctx(ActionCtx::DeleteAccount(password))) => match &profile.auth {
-            Some(_) => Effects::one(delete_account::<E>(password)).unchanged(),
+        Msg::Action(Action::Ctx(ActionCtx::DeleteAccount(password))) => match profile.auth_key() {
+            Some(auth_key) => Effects::one(delete_account::<E>(auth_key, password)).unchanged(),
             _ => Effects::msg(Msg::Event(Event::Error {
                 error: CtxError::from(OtherError::UserNotLoggedIn),
                 source: Box::new(Event::UserAccountDeleted { uid: profile.uid() }),
@@ -390,7 +390,10 @@ pub fn update_profile<E: Env + 'static>(
                 }
             }
         }
-        Msg::Internal(Internal::DeleteAccountAPIResult(_, result)) => match result {
+        Msg::Internal(Internal::DeleteAccountAPIResult(
+            APIRequest::DeleteAccount { auth_key, .. },
+            result,
+        )) if profile.auth_key() == Some(auth_key) => match result {
             Ok(_) => Effects::msg(Msg::Internal(Internal::Logout)).unchanged(),
             Err(error) => Effects::msg(Msg::Event(Event::Error {
                 error: error.to_owned(),
@@ -507,8 +510,9 @@ fn push_profile_to_storage<E: Env + 'static>(profile: &Profile) -> Effect {
     .into()
 }
 
-fn delete_account<E: Env + 'static>(password: &String) -> Effect {
+fn delete_account<E: Env + 'static>(auth_key: &AuthKey, password: &String) -> Effect {
     let request = APIRequest::DeleteAccount {
+        auth_key: auth_key.to_owned(),
         password: password.to_owned(),
     };
     EffectFuture::Concurrent(
--- a/src/models/ctx/update_profile.rs
+++ b/src/models/ctx/update_profile.rs
@@ -11,7 +11,7 @@ use crate::types::addon::Descriptor;
 use crate::types::api::{
     fetch_api, APIError, APIRequest, APIResult, CollectionResponse, SuccessResponse,
 };
-use crate::types::profile::{Auth, AuthKey, Profile, Settings, User};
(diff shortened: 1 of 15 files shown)


src/models/ctx/update_profile.rs
```diff
@@ -491,6 +510,24 @@ fn push_profile_to_storage<E: Env + 'static>(profile: &Profile) -> Effect {
     .into()
 }
 
+fn delete_account<E: Env + 'static>(auth_key: &AuthKey, password: &String) -> Effect {
```
elpiel:
```suggestion
fn delete_account<E: Env + 'static>(auth_key: &AuthKey, password: &str) -> Effect {
```

src/runtime/msg/action.rs
```diff
@@ -34,6 +34,7 @@ use crate::{
 pub enum ActionCtx {
     Authenticate(AuthRequest),
     Logout,
+    DeleteAccount(String),
```
elpiel: for fields that don't have a type, I'd suggest to add a field instead of just the type:
`DeleteAccount { password: String }`

src/unit_tests/ctx/delete_account.rs
```diff
@@ -0,0 +1,158 @@
...
+use crate::types::library::LibraryBucket;
+use crate::types::notifications::NotificationsBucket;
+use crate::types::profile::{Auth, AuthKey, GDPRConsent, Profile, User};
+use crate::types::search_history::SearchHistoryBucket;
+use crate::types::server_urls::ServerUrlsBucket;
+use crate::types::streams::StreamsBucket;
+use crate::types::True;
+use crate::unit_tests::{
+    default_fetch_handler, Request, TestEnv, FETCH_HANDLER, REQUESTS, STORAGE,
+};
```
elpiel: can you please group imports in single import for new code, thanks!
 also external crate imports usually go above local ones, so futures and std should be above.

I like to separate `std/core` from the rest of the imports as well.

src/unit_tests/ctx/delete_account.rs
```diff
@@ -0,0 +1,158 @@
...
+    assert_eq!(
+        REQUESTS.read().unwrap().get(1).unwrap().to_owned(),
+        Request {
+            url: "https://api.strem.io/api/logout".to_owned(),
+            method: "POST".to_owned(),
+            body: "{\"type\":\"Logout\",\"authKey\":\"auth_key\"}".to_owned(),
+            ..Default::default()
+        },
+        "Logout request has been sent"
+    );
```
elpiel: Just to double check:
The API will still be able to Log us out using the same auth key if we first delete the account?

src/models/ctx/update_profile.rs
```diff
@@ -30,6 +30,14 @@ pub fn update_profile<E: Env + 'static>(
                 Effects::none().unchanged()
             }
         }
+        Msg::Action(Action::Ctx(ActionCtx::DeleteAccount(password))) => match profile.auth_key() {
+            Some(auth_key) => Effects::one(delete_account::<E>(auth_key, password)).unchanged(),
```
elpiel, line 34: I think that Niki would add the UID to the action here as well as a guard in case that Delete ACcount was triggered with different account. 
I think it should be safe in this case as the API will return bad response if password doesn't mathc anyway.

src/unit_tests/ctx/delete_account.rs
tymmesyde: This is a downside from this: https://github.com/Stremio/stremio-core/pull/753/files#diff-5e5b6a800019c0be2c32e04c71c27cda593eaef5d4b50e2a6772eb6e8cffac88R397
It will dispatch the `Internal:Logout` message so that every update functions can remove the user data
But this message is also handled in the ctx model which trigger a delete session / logout request
https://github.com/Stremio/stremio-core/blob/354f6b1734e37a44ea23ed99234d22b90daa7bfa/src/models/ctx/ctx.rs#L119

src/models/ctx/update_profile.rs
tymmesyde: There is a guard there: https://github.com/Stremio/stremio-core/pull/753/files#diff-5e5b6a800019c0be2c32e04c71c27cda593eaef5d4b50e2a6772eb6e8cffac88R396
It's checking if the auth_key from the current profile is the same as the one that was sent in the request to prevent this

src/runtime/msg/action.rs
tymmesyde: I can create a type for Password which would just be an alias for String if you want? Similar to LibraryItemId
Using a struct for this is not very useful as we don't have other arguments to send and result in more cumbersome API

src/unit_tests/ctx/delete_account.rs
tymmesyde: I grouped them, tho the ordering is handled by `cargo fmt`

elpiel (approved):
LGTM!

PR #755, opened 2024-12-23
fix(player): reset library item time offset if selected video don't match state

tymmesyde:
When library item and selected is available, if the selected requested video_id doesn't match the state of the stored library item, update the item in the library to reset its timeOffset to 0


kKaskak (approved):
tested with web and it works ✅

jaruba (approved):

jaruba:
tested, works fine

PR #756, opened 2024-12-23
fix(core-web): return stream progress only for corresponding videoId

tymmesyde:


kKaskak (approved):
LGTM, works correctly tested with web locally

PR #763, opened 2025-01-08
feat(stremio-core-web): MetaDetails - add last used stream flag to Streams

elpiel:
Marks the latest streams previously watched on the device as `last used` for stremio-core-web MetaDetails model

- renames core's field to fit better as the name was misleading
- adds field to stremio-core-web

### DIFF
--- a/stremio-core-web/src/model/serialize_meta_details.rs
+++ b/stremio-core-web/src/model/serialize_meta_details.rs
@@ -44,9 +44,11 @@ mod model {
         /// Watch progress percentage
         pub progress: Option<f64>,
         pub deep_links: StreamDeepLinks,
-        /// Whether or not this is the suggested by core stream
-        /// Only 1 stream can be a suggested one!
-        pub suggested: Option<bool>,
+        /// Whether or not this is a stream the user has already played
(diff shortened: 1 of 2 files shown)


stremio-core-web/src/model/serialize_meta_details.rs
```diff
@@ -41,9 +41,14 @@ mod model {
...
         #[serde(flatten)]
         pub stream: &'a stremio_core::types::resource::Stream,
-        // Watch progress percentage
+        /// Watch progress percentage
         pub progress: Option<f64>,
         pub deep_links: StreamDeepLinks,
+        /// Whether or not this is a stream the user has already played
+        /// Only 1 stream can be the last used one!
+        ///
+        /// Find out more about how we select it form the StreamsBucket in the core's model.
```
elpiel:
```suggestion
        /// Find out more about how we select it from the StreamsBucket in the core's model.
```

kKaskak:
```suggestion
        /// Find out more about how we select it from the StreamsBucket in the core's model.
```

kKaskak (approved):

PR #788, opened 2025-02-28
feat(Auth): impl apple token auth

kKaskak:


src/types/profile/user.rs
```diff
@@ -99,6 +99,9 @@ pub struct User {
     pub fb_id: Option<String>,
     #[serde(default)]
     #[serde_as(deserialize_as = "DefaultOnNull<NoneAsEmptyString>")]
+    pub apple_id: Option<String>,
```
elpiel, line 102: We should use an enum to represent the different authentication ids.

kKaskak: Will do this

elpiel (approved):
LGTM

PR #796, opened 2025-03-26
fix(core): stremio-serde-hex fix for panics on empty input

elpiel:
tests have been added to confirm that panics will not occur from serde-hex


kKaskak (approved):
LGTM
Screenshot 2025-03-31 at 10 55 26
Screenshot 2025-03-31 at 10 55 59

PR #799, opened 2025-04-07
feat(User): Impl trakt expired event and hasTrakt field

kKaskak:
related: https://github.com/Stremio/stremio-web/pull/884


elpiel (changes requested):
We are working on a different solution for refreshing the trakt token.

PR #804, opened 2025-04-15
feat(Subtitles): Changed id format

ivelinmegdanov:
Linked to: https://github.com/Stremio/stremio-web/pull/894

Closes [Issue #403](https://github.com/Stremio/stremio-tasks/issues/403)

### DIFF
--- a/stremio-core-web/src/model/serialize_player.rs
+++ b/stremio-core-web/src/model/serialize_player.rs
@@ -206,7 +206,7 @@ pub fn serialize_player<E: stremio_core::runtime::Env + 'static>(
                     // renames the subtitle id to avoid conflicts
                     .map(move |(position, subtitles)| model::Subtitles {
                         subtitles,
-                        id: format!("{}_{}", addon.transport_url, position),
+                        id: format!("Sub {} ID: {}", position, subtitles.id.clone()),
                         origin: &addon.manifest.name,
                     })
             })


stremio-core-web/src/model/serialize_player.rs
```diff
@@ -206,7 +206,7 @@ pub fn serialize_player<E: stremio_core::runtime::Env + 'static>(
                     // renames the subtitle id to avoid conflicts
                     .map(move |(position, subtitles)| model::Subtitles {
                         subtitles,
-                        id: format!("{}_{}", addon.transport_url, position),
+                        id: format!("Sub {} ID: {}", position, subtitles.id.clone()),
```
tymmesyde, line 209: What is the purpose of this change?

ivelinmegdanov: The purpose is to include the original subtitles.id (which gives us meaningful info like the actual source or filename explained in the issue) i dont know why it was changed to be the transport_url in the first place. But also we still keep the position to avoid potential conflicts when multiple subtitles share the same ID.

tymmesyde: The purpose of the id of a subtitles is just to provide a unique id and is used with stremio-video for track selection
Since it's up to the addons dev to set this id, we use the transport url to make sure that two addons don't use the same id
The id that is set by the addon could be anything like random numbers, it doesn't necessarily contains the filename

Here are the specs for the subtitles object:
https://github.com/Stremio/stremio-addon-sdk/blob/master/docs/api/responses/subtitles.md
I suggest that we had a field here like `fileName` so that we and the addon devs can pass the name of the file

kKaskak:
outdated

PR #811, opened 2025-04-28
Fixing a deep link that wasn't working due to an unescaped string

Emahhh:
When building the IINA deep link, the URL wasn't escaped correctly, so some URLs weren't working because IINA wasn't reading the complete URL. Now the URL should be escaped correctly.
The same bug could be present in other deep link integrations.


elpiel:
Hello and thank you for opening the PR.
This PR is a duplicate and strip down version of the fix introduced in this PR: #787 
As you will see there are other x-callback schemas that needed fixing.

PR #838, opened 2025-07-30
feat(settings): save playback speed and volume

Aztup:


src/types/profile/settings.rs
```diff
@@ -43,6 +43,8 @@ pub struct Settings {
     pub streaming_server_warning_dismissed: Option<DateTime<Utc>>,
     pub server_in_foreground: bool,
     pub send_crash_reports: bool,
+    pub volume: u8,
+    pub playback_speed: i16
```
tymmesyde, line 47: this should be `f32` and default value set to `1.0`

Aztup: I think volume should probably stay as `u8` just like `subtitles_opacity`

Aztup: I also forgot the mention but the reason it's an `i16` is because I would need to modify a lot of files to remove the `Eq` trait so should I keep it as `i16` instead?

kKaskak:
outdated

PR #839, opened 2025-08-07
Feat/pull user from api via token

AlvinHV:


src/runtime/msg/action.rs
```diff
@@ -55,7 +55,14 @@ pub enum ActionCtx {
     DismissNotificationItem(MetaItemId),
     ClearSearchHistory,
     PushUserToAPI,
-    PullUserFromAPI,
+    PullUserFromAPI {
```
kKaskak, line 75: wouldnt this require changes in each app since we are changing the action from enum tuple variant to enum struct variant

kKaskak (changes requested):

src/runtime/msg/action.rs
elpiel: Yes it does.
Updating core can have breaking  changes and I don't see the reason to make a separate action for this as it should be 1 action that handles this.
Otherwise we have duplicates just for the sake of not changing the apps.


+ with default, it's not mandatory to send the field, it will be defaulted.
Need to check if changing the json passed as action by web is required though.
The rest of the apps have type checks.

elpiel (approved):

src/runtime/msg/action.rs
elpiel: Sadly `args` is reaquired even though the field is `option`al and `default`ed.
This would require a change in web, however, it's a good trade-off for reusing all the code of the action

elpiel (approved):

PR #843, opened 2025-08-13
fix(library_item): include removed library items from continue watching

tymmesyde:


elpiel (approved):
As discussed with @jaruba , we should include removed and temporary library items.

PR #860, opened 2025-10-06
fix: lib. item - is_in_cw fn - re-added !self.removed check

elpiel:


AlvinHV:
Tested on iOS/tvOS app and CW behaves expected

AlvinHV (approved):

PR #917, opened 2026-01-13
feat: external player callback tvos

AlvinHV:


AlvinHV:
@elpiel there may be problem in web. In web version infuse will try opening stremio app but not sure how to make callback only work on native app. I tried passing player data to infuse but it cut off after third slash.

kKaskak (approved):

PR #924, opened 2026-01-30
fix: Enable ASS/SSA subtitle styling by default

### LINKED ISSUE #910: [Bug] ASS/SSA subtitles styles ignored in newer versions
In the older stable version 4.4 subtitles with custom styles (ASS/SSA) were rendered correctly, with the correct fonts, colors, and specific screen positioning. In the current version, these subtitles are stripped of their formatting and displayed as plain white text (similar to SRT). 

Image
^This is with stremio 5.x

Image
^This is with stremio 4.4

Am i missing something in the settings to display the correct font?

anikettuli:
Fixes #910 - ASS/SSA subtitles styles were being ignored in newer versions.

Changed the default value of ass_subtitles_styling from false to true to match v4.4 behavior where subtitle styles (fonts, colors, positioning) were correctly preserved.

Changes:
- Set ass_subtitles_styling default to true in Settings
- Updated migration v21 to set assSubtitlesStyling to true
- Updated related test expectations

Testing:
- All 194 existing tests pass.

### DIFF
--- a/src/runtime/env.rs
+++ b/src/runtime/env.rs
@@ -758,7 +758,7 @@ fn migrate_storage_schema_to_v21<E: Env>() -> TryEnvFuture<()> {
                 Some(settings) => {
                     settings.insert(
                         "assSubtitlesStyling".to_owned(),
-                        serde_json::Value::Bool(false),
+                        serde_json::Value::Bool(true),
                     );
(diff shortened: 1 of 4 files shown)


kKaskak:
im ok with this change

tymmesyde:
Hi, we purposefully disabled it by default because it causes some issues with certain subtitles

PR #932, opened 2026-02-14
feat(core-web): expose pagination loading state in Discover serializer

klnstprx:
## Summary
- Add `loadingNextPage` boolean to the serialized Discover selectable state
- True when there are 2+ catalog pages and the last page is still loading (pagination in progress)
- Gives the frontend an authoritative signal to show a loading indicator instead of relying on fragile local React state


kKaskak:
see https://github.com/Stremio/stremio-web/pull/1133

PR #968, opened 2026-04-20
chore: remove unmaintained boolinator dep

anikettuli:
## Summary

Remove the `boolinator` crate (unmaintained since 2016 — last release August 2016) from the workspace. Replace its `Boolinator::as_option` / `as_some` extension methods on `bool` with stdlib equivalents (`bool::then`, `bool::then_some`).

## Why this change

- `boolinator` is unmaintained (no commits in 9+ years); `rustsec` has flagged it as a stale dependency in the past.
- `bool::then` (stable since Rust 1.50, May 2021) and `bool::then_some` (stable since Rust 1.62, June 2022) cover every call site. No third-party dep needed.
- Shrinks the dependency graph by one crate; `Cargo.lock` no longer resolves `boolinator 2.4.0`.

## Effect

- **One fewer transitive dependency**: slightly faster clean builds, fewer vectors for supply-chain staleness alerts.
- **No behavior change**: the replacements are semantically identical. `bool::then(F)` returns `Some(F())` when true; `as_option(x)` returned `Some(x)` when true — equivalent except `then` accepts a closure so side-effectful arguments only execute on the truthy branch (never the case here).
- **No public API / no serialized-shape / no wire-format changes.** This is an internal refactor only.

## Call sites updated

- `src/models/catalog_with_filters.rs` — `(!extra_prop.is_required).as_option().map(|_| SelectableExtraOption { .. })` → `(!extra_prop.is_required).then(|| SelectableExtraOption { .. })`
- `src/types/resource/stream.rs` (`Stream::youtube`) — `cond.as_option().and_then(|_| split)` → `cond.then(|| split).flatten()`
- `stremio-core-web/src/model/serialize_discover.rs` — `(!discover.catalog.is_empty()).as_option().map(|_| ..)` → `.then(|| ..)`

Manifest changes:
- `Cargo.toml` — drop `boolinator = "2.4"`
- `stremio-core-web/Cargo.toml` — drop `boolinator = "2.4.*"`
- `Cargo.lock` — regenerated (removes the `boolinator` entry)

## Test plan

Verified locally on Windows 11 with `rustc 1.95.0 stable-x86_64-pc-windows-gnu`:

- [x] `cargo clippy -p stremio-core --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` — clean (fixup commit `082e3f63f` applies the formatting that `cargo fmt` requested after the `.then(..)` rewrite)

Full-workspace build (including `stremio-core-web`) needs the `wasm-bindgen` upgrade in #969 because `wasm-bindgen 0.2.78` is incompatible with current Rust.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -102,12 +102,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "boolinator"
-version = "2.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cfa8873f51c92e232f9bac4065cddef41b714152812bfc5f7672ba16d6ef8cd9"
-
-[[package]]
 name = "bumpalo"
 version = "3.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1107,7 +1101,6 @@ dependencies = [
  "anyhow",
  "assert_matches",
  "base64 0.22.1",
- "boolinator",
  "chrono",
  "data-encoding",
  "derivative",
@@ -1154,7 +1147,6 @@ dependencies = [
 name = "stremio-core-web"
 version = "0.51.1"
 dependencies = [
- "boolinator",
  "chrono",
  "console_error_panic_hook",
  "either",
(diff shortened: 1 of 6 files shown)


anikettuli:
Superseded by #973 — the five dead-weight/MSRV modernizations are bundled there to avoid five near-identical Cargo.toml / Cargo.lock merges. Each change remains a distinct commit for piecewise review.

PR #973, opened 2026-04-20
chore: workspace modernization — MSRV 1.85, drop 4 unmaintained deps

anikettuli:
## Summary

Bundles five contract-preserving modernization steps into one PR:

1. **MSRV 1.85** — aligns declared `rust-version` with the CI matrix across all three workspace crates.
2. **Remove `boolinator`** — unmaintained since 2016; covered by stdlib `bool::then` / `then_some`.
3. **Replace `tokio-current-thread = "=0.2.0-alpha.1"`** — a pre-1.0 alpha, with modern `tokio::task::LocalSet`.
4. **Migrate `once_cell::sync::Lazy` → `std::sync::LazyLock`** — drop-in replacement since Rust 1.80.
5. **Remove `enclose` macro** — the `enclose!` convenience is replaced by plain `let x = x.clone(); move ||` blocks (a no-dep idiom since Rust 2021).

These are independent conceptually but share a common theme ("modernize the dependency floor without touching runtime behavior") and overlap heavily in manifest edits, so bundling avoids five near-identical `Cargo.toml` / `Cargo.lock` merge conflicts.

Each step lands as its own commit so reviewers can read, test, or revert piecewise:

```
f6a7b6d33 build: bump workspace MSRV from 1.77 to 1.85
bc370a59c chore: remove unmaintained boolinator dep
bda1377bc test: replace tokio-current-thread alpha with tokio LocalSet
aa377c8cd chore: migrate once_cell::sync::Lazy to std::sync::LazyLock
d6f597835 chore: remove enclose macro; use explicit clone-before-move
```

## Why these changes

- **`boolinator`**: last release August 2016. Covered by `bool::then` (Rust 1.50) and `bool::then_some` (1.62). Keeping it is pure supply-chain surface area for zero benefit.
- **`tokio-current-thread`**: pinned at a pre-1.0 *alpha* (`=0.2.0-alpha.1`) from 2018. The only consumer is the test harness; modern tokio has `LocalSet` that does the exact same `!Send` single-threaded spawn-and-drain.
- **`once_cell`**: `LazyLock` shipped in Rust 1.80. Using the stdlib version removes a third-party crate from every build.
- **`enclose`**: the `enclose!` macro is a pre-Rust-2021 ergonomic shortcut. Rust 2021's disjoint-capture rules plus plain `let x = x.clone();` expand to the exact same code — no macro needed.
[...]

PR #969, opened 2026-04-20
build: upgrade wasm-bindgen 0.2.78 → 0.2.118 and WASM deps

anikettuli:
## Summary

Unpins and upgrades `wasm-bindgen` from the 4-year-old hard-pinned `=0.2.78` to `0.2.118`, aligns the rest of the WASM dependency stack, and unblocks `wasm-pack build` on current Rust toolchains.

## Why this change

The previous `wasm-bindgen = "=0.2.78"` pin was incompatible with any Rust newer than ~1.77, blocking the entire workspace on a specific historical toolchain. Running `cargo build` on current stable Rust produces:

```
error: older versions of the `wasm-bindgen` crate are incompatible with current
versions of Rust; please update to `wasm-bindgen` v0.2.88
```

A corresponding pin in `stremio-core-web/rust-toolchain.toml` forced Rust 1.77 specifically to work around this. That workaround paid the cost of stalling contributors on an old toolchain and masking real warnings from newer clippy/rustc.

## Effect

- **Developers can build the project on current stable Rust** (1.85+).
- **`wasm-pack build stremio-core-web --release --target web` produces a working `.wasm` binary** on today's toolchain (measured: 5,045,686 bytes on Rust 1.85).
- **~40 versions of upstream fixes** become available (macro hygiene, Promise handling, JsValue API refinements, ES modules support).
- **clippy -D warnings is reachable again.** One pre-existing dead-code warning (`SelectablePage` DTO) was previously masked by the build failure; it is surfaced now and annotated (see below).
- **No public `#[wasm_bindgen]` export renamed or re-shaped.** `initialize_runtime`, `dispatch`, `get_state`, `get_debug_state`, `analytics`, `encode_stream`, `decode_stream` — all byte-for-byte identical.
- **No `Msg` / `Action` / `Event` variant changes, no deep-link grammar changes, no JSON field renames.**

### `stremio-core-web/Cargo.toml`
- `wasm-bindgen = "=0.2.78"` → `"0.2.118"` (unpinned — let patches apply automatically).
- Dropped the deprecated `serde-serialize` feature on `wasm-bindgen`. JSON interop already flows through `gloo_utils::format::JsValueSerdeExt`, so no runtime behavior changes.
- Normalized `wasm-bindgen-futures`, `js-sys`, `web-sys`, `gloo-utils`, `tracing-wasm`, `console_error_panic_hook`, `wasm-bindgen-test` to caret ranges so Cargo picks compatible patches automatically.

[...]

PR #974, opened 2026-04-20
build: unblock cargo update + stremio-watched-bitfield base64 0.13 → 0.22

anikettuli:
## Summary

Two commits that together unblock routine `cargo update` from the workspace without spurious CI breakage, and retire the last remaining base64 major 0.x in the workspace:

1. **`test:`** Round-trip deep-link URL assertions instead of byte-comparing. Fixes 12 tests that asserted exact byte equality of a zlib-compressed player URL — which broke on every `miniz_oxide` / `flate2` output change despite no semantic regression.
2. **`build:`** Run `cargo update` to pull in latest compatible patches, and migrate `stremio-watched-bitfield` from `base64 = "0.13.0"` (Aug 2021) to `0.22`.

### Brittle tests

12 tests in `src/unit_tests/deep_links/` (across `stream_deep_links.rs`, `meta_item_deep_links.rs`, `video_deep_links.rs`) asserted that a `StreamDeepLinks::player` URL string equals a specific byte sequence. The first path segment of that URL is a `zlib-compressed → base64 → percent-encoded` JSON payload produced by `Stream::encode`.

Different versions of `miniz_oxide` (`flate2`'s compression backend) emit the same decompressed bytes through **different compressed byte sequences**. So:
- No semantic regression — the URL still decodes to the exact same `Stream`.
- But `assert_eq!(sdl.player, "stremio:///player/eAEB...")` fails.

This made every routine `cargo update` break 12 tests. The fix: compare the **decoded** stream semantically, and keep the URL-grammar suffix under literal assertion.

### Unblocked `cargo update`

With the brittle tests fixed, we run `cargo update` and pick up ~2 years of accumulated upstream patches:
- `regex`, `serde`, `tokio`, `tracing`, `url`, `derive_more`, `itertools`, etc. — 50+ deps updated.
- Two deps **pinned back** to their last MSRV-1.85-compatible release because their latest requires Rust 1.88:
  - `serde_with` / `serde_with_macros` at `3.12.0` (3.18 needs 1.88)
  - `darling` at `0.20.11` (0.23 needs 1.88)
  These pins come off the moment the workspace MSRV rises past 1.88.

### `stremio-watched-bitfield` base64 upgrade

`stremio-watched-bitfield` was still on `base64 = "0.13.0"` (August 2021). The free-function `decode(s)` / `encode(s)` API was removed in `base64 0.21` in favor of the `Engine` trait. Migrating:

```rust
// before (base64 0.13):
use base64::{decode, encode};
let bytes = decode(s)?;
let s = encode(bytes);

// after (base64 0.22):
use base64::{engine::general_purpose::STANDARD as BASE64, Engine};
let bytes = BASE64.decode(s)?;
let s = BASE64.encode(bytes);
```

Wire-format is identical — `STANDARD` matches `0.13`'s default alphabet, so serialized `BitField8` strings round-trip byte-for-byte between old and new versions. All 7 existing watched-bitfield tests confirm this.

## Effect

- **Routine `cargo update` now lands cleanly** without any spurious test failures. Dependency drift becomes a no-drama operation.
- **Every workspace crate on current base64** (0.22 across root + stremio-core-web + stremio-watched-bitfield). The last 3-year-old major-version dep floor is gone.
- **50+ transitive patches** applied (bugfixes, small perf improvements, supply-chain freshness).
- **No runtime behavior change.** Wire formats, deep-link grammar, and JSON shapes are untouched.

## Public contract

Zero changes to:
- Storage keys at [src/constants.rs:10](src/constants.rs)
- WASM exports at [stremio-core-web/src/stremio_core_web.rs:83-348](stremio-core-web/src/stremio_core_web.rs)
- `Msg` / `Action` / `Event` variants
- Deep-link URL grammar (the tests still verify the URL prefix + suffix structure literally; only the opaque encoded-payload segment is checked semantically now)
- Serialized JSON field names
- `BitField8` wire format (verified by existing round-trip tests)

## Test plan

Verified locally (Rust 1.85.1 `stable-x86_64-pc-windows-gnu`):

- [x] `cargo test -p stremio-core --lib unit_tests::deep_links` — **41 passed**
- [x] `cargo test -p stremio-watched-bitfield` — **7 passed, 0 failed**
- [x] `cargo clippy --all --no-deps -- -D warnings` — clean

## Dependency on other PRs

- **Independent** of #967 / #969 / #973. Merges in any order.
- The MSRV pins on `serde_with` / `darling` reference MSRV 1.85 which is already on stable; if MSRV ever bumps past 1.88, those pins come off in a one-line follow-up.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,34 +3,19 @@
 version = 3
 
 [[package]]
-name = "addr2line"
-version = "0.24.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f5fb1d8e4442bd405fdfd1dacb42792696b0cf9cb15882e5d097b742a676d375"
-dependencies = [
- "gimli",
-]
-
-[[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
+checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
 dependencies = [
  "memchr",
 ]
-
-[[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
 
 [[package]]
 name = "android_system_properties"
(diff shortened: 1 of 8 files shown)


kKaskak:
this prs adds no value again, no bug fix no feature addition

anikettuli:
the test fix isn't code churn for its own sake as it specifically makes future cargo update runs not break CI on miniz_oxide bumps, and the base64 0.13 dep is 3+ years out of date.

PR #975, opened 2026-04-20
test: insta snapshot tripwire for persisted-storage JSON shapes

anikettuli:
## Summary

Adds a new `src/unit_tests/snapshots.rs` module that uses the `insta` crate to **freeze the serialized JSON shape** of every bucket that `stremio-core` writes to persistent storage. Each storage key listed in [`src/constants.rs`](src/constants.rs) now has a committed `.snap` fixture; any schema or field-name change to one of these types produces a visible `cargo insta review` diff that cannot accidentally land.

## Why this change

Storage keys (`profile`, `library`, `library_recent`, `streams`, `search_history`, `streaming_server_urls`, `notifications`, `calendar`, `dismissed_events`) are **installed on end-user devices**. A silent rename of a serialized field — say a typo fix on a variant, or a `#[serde(rename)]` added during a refactor — breaks every existing install with no compiler warning and no user-visible cue until someone tries to load their saved state.

The existing coverage in `src/unit_tests/serde/` tests round-trips for individual types, but has no complete tripwire for the composite bucket shapes. If someone adds a field to `Profile` that defaults to `serde_json::Value::Null` on deserialization but changes the serialized output, the round-trip tests still pass. The snapshot suite catches that kind of drift.

## Effect

- **Any field rename, shape change, or default-value change** on a storage-bound type produces an `insta` diff in CI. Contributors can't accidentally break existing user data.
- Updating a snapshot intentionally is still fast: `cargo insta review` → inspect the diff → accept. The expected pairing with a migration step in [`src/runtime/env.rs`](src/runtime/env.rs) is now a visible, enforceable reviewer checkbox rather than an unwritten rule.
- Zero runtime behavior change. Tests-only addition.

## Scope chosen deliberately

**In scope (this PR):** storage bucket types — the persistence contract.
- `Profile` (default state, including the bundled official addons)
- `LibraryBucket` (empty) — covers both `library` and `library_recent`
- `StreamsBucket`, `SearchHistoryBucket`, `ServerUrlsBucket`, `NotificationsBucket`, `DismissedEventsBucket` (empty)

**Out of scope (not this PR):**
- `Msg` / `Action` / `Event` JSON — these are dispatch-internal, not persisted. Breaking them affects live clients (where the on-the-wire contract matters for a running app session), but doesn't corrupt installed user data. The existing `src/unit_tests/serde/` suite covers this class at the type level.
- Deep-link URLs — already covered by existing `src/unit_tests/deep_links/*.rs` tests (now with a round-trip helper for the flate2-encoded player URL, see #974).

## Files changed

- `src/unit_tests/snapshots.rs` — new, 7 tests.
- `src/unit_tests/snapshots/*.snap` — 7 committed fixtures.
- `src/unit_tests/mod.rs` — register the module.
- `Cargo.toml` — add `insta = { version = "1", features = ["json"] }` to `[dev-dependencies]`.

## How to review a snapshot change in a future PR

When a legitimate schema change lands, the author runs:

```
cargo insta review
```

which shows each new/changed snapshot side-by-side with the previous committed version. The author accepts the diff and commits the updated `.snap` file in the same PR as the code change. Reviewers see the `.snap` diff alongside the type change and can verify:

1. Is the rename/addition intentional?
2. Is there a migration step in `src/runtime/env.rs` for it?
3. Does the migration have its own test fixture?

## Test plan

Verified locally (Rust 1.85.1 `stable-x86_64-pc-windows-gnu`):

- [x] `cargo test -p stremio-core --lib` — **209 passed, 0 failed** (202 original + 7 new snapshot tests)
- [x] `cargo clippy --all --no-deps -- -D warnings` — clean

## Dependency on other PRs

- **Independent** of #967 / #969 / #973 / #974. Merges in any order.
- Lives under `[dev-dependencies]` only, so no effect on published crate graph.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,12 +27,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
-
-[[package]]
 name = "android_system_properties"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -154,17 +148,28 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "chrono"
-version = "0.4.38"
+version = "0.4.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a21f936df1771bf62b77f047b726c4625ff2e8aa607c01ec06e5a05bd8463401"
+checksum = "c673075a2e0e5f4a1dde27ce9dee1ea4558c7ffe648f576438a20ca1d2acc4b0"
 dependencies = [
- "android-tzdata",
  "iana-time-zone",
  "js-sys",
  "num-traits",
  "serde",
  "wasm-bindgen",
- "windows-targets",
+ "windows-link",
+]
+
+[[package]]
+name = "console"
+version = "0.15.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "054ccb5b10f9f2cbf51eb355ca1d05c2d279ce1804688d0db74b4733a5aeafd8"
+dependencies = [
+ "encode_unicode",
+ "libc",
+ "once_cell",
+ "windows-sys",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


kKaskak:
i think the real fix for this would be to ignore the fields that cannot be parsed etc instead of just guarding, example now users cannot downgrade core if they would like to right? the app would break adding tests here does absolutely nothing 
Ui handles that by prompting to clear data for now

anikettuli:
Makes sense — if the UI-level "clear data" flow is the intended backstop, a CI-only tripwire is friction without matching value. Appreciate you taking the time to explain rather than just closing. I'll keep the reasoning in mind for future contributions.

PR #980, opened 2026-05-14
Feat/add gpuVideoProcessing option

Aztup:
Add `nvidiaVideoProcessing` field to support RTX VSR/HDR as a toggle in stremio.
This setting is disabled by default.

I was also thinking maybe renaming it to something like `gpuVideoProcessing` could make a little bit more 'feature proof'


kKaskak:
ok

kKaskak (approved):

PR #988, opened 2026-05-25
fix(stremio-watched-bitfield): store bit length in BitField8::new

### LINKED ISSUE #987: Wrong last-watched anchor in serialized WatchedField (BitField8::new stores byte length)
**Description**

`BitField8::new` stores the number of *bytes* in the `length` field instead of the number of *bits*:

```rust
pub fn new(length: usize) -> BitField8 {
    let length = (length as f64 / 8.0).ceil() as usize; // byte count
    BitField8 {
        length,
        values: vec![0; length],
    }
}
```

Everywhere else in the crate `length` is the bit count:

- `new_with_values` sets `length = values.len() * 8`.
- `set` sets `self.length = self.values.len() * 8` when the backing buffer grows.
- `last_index_of` iterates `0..self.length`, treating it as bit indices.
- The `BitField8` round-trip tests assert `length == 16` for two bytes.

Because `new` sets `length` to `ceil(bits / 8)`, a bitfield created with `new` reports a `length` that is ~8x too small, and `last_index_of(..)` only scans the first `ceil(n / 8)` bit positions.

This surfaces in the watched-state serialization. `WatchedField`'s anchor (`anchor_video` / `anchor_length`) is derived from `last_index_of(true)` in both `From<WatchedBitField> for WatchedField` and `WatchedBitField`'s `Display`. The anchor is documented as the last watched video id. When a `WatchedBitField` is built through `new` — the `construct_from_array` fallback used for library items without prior watched state, and the resize buffer in `construct_with_videos` — and a video past the first `ceil(n / 8)` bits is marked watched, `last_index_of` fails to find it and the anchor collapses to video index 0 (or `undefined`) with `anchor_length = 1`.

Concretely, in the player, watching an episode to the completion threshold on a series with no prior watched state runs `construct_from_array(vec![], video_ids)` then `set_video(..)` then `WatchedField::from(..)`. For a 20-episode series, marking episode 16 (index 15) watched serializes an anchor of `tt..:1:1` instead of `tt..:16:16`. The serialized `WatchedField` is part of the library format synced through the API and shared with other Stremio clients, so a wrong anchor can misalign or drop watched state when the video list later changes.

This is distinct from the off-by-one in `last_index_of` fixed in #781; that change corrected the scan range, but the `length` value fed to it from `new` is still wrong.

**Expected behavior**

`BitField8::new(n)` creates a bitfield whose `length` is `n` bits (consistent with `new_with_values`), so `last_index_of` scans every bit position and `WatchedField` records the actually-last-watched video.

**Actual behavior**

`BitField8::new(n)` sets `length = ceil(n / 8)`, so `last_index_of` under-scans and `WatchedField` / `WatchedBitField` serialize an incorrect anchor for any watched video beyond the first `ceil(n / 8)` positions.

**Implementation notes**

Store the bit length in `length` and size the backing `Vec<u8>` to `ceil(length / 8)` bytes. Add regression coverage for `BitField8::new(..)` length and `last_index_of`, and for the `WatchedField` anchor produced via the `construct_from_array` + `set_video` path.

officialasishkumar:
## Summary

`BitField8::new` stored the byte count (`ceil(length / 8)`) in the `length` field, while `new_with_values`, `set` and `last_index_of` all treat `length` as the number of bits. Because of this, `last_index_of` only scanned the first `ceil(n / 8)` bit positions of any bitfield created through `new`.

The visible effect is a wrong `WatchedField` anchor. The anchor (`anchor_video` / `anchor_length`, documented as the last watched video) is derived from `last_index_of(true)`. A `WatchedBitField` built via `new` (the `construct_from_array` fallback for library items with no prior watched state, and the resize buffer in `construct_with_videos`) serialized an anchor that collapsed to video index 0 with length 1 once the last watched video was past the first `ceil(n / 8)` bits. In the player this happens when watching an episode to the completion threshold on a series with no prior watched state. For a 20 episode series, marking episode 16 watched produced `..:1:1` instead of `..:16:16`. As the serialized `WatchedField` is synced through the API and shared with other Stremio clients, the wrong anchor can misalign or drop watched state when the video list changes.

This is separate from the `last_index_of` off by one fixed in #781, which corrected the scan range but not the `length` value passed into it.

## Fix

Keep `length` in bits and size the backing `Vec<u8>` to `ceil(length / 8)` bytes.

## Tests

- `bitfield8::tests::new_uses_bit_length`: `BitField8::new(n).length == n` and `last_index_of` finds a bit past the first byte.
- `watched_bitfield::tests::watched_field_anchor_points_to_last_watched_video`: the `construct_from_array` + `set_video` flow (as used by the player) produces an anchor pointing at the actually watched video.

Verified locally: `stremio-watched-bitfield` (9 passed) and `stremio-core` (202 passed, single threaded) test suites are green, along with rustfmt and clippy.

Closes #987

### DIFF
--- a/stremio-watched-bitfield/src/bitfield8.rs
+++ b/stremio-watched-bitfield/src/bitfield8.rs
@@ -15,9 +15,6 @@ pub struct BitField8 {
 
 impl BitField8 {
     pub fn new(length: usize) -> BitField8 {
-        // `length` is the number of bits (videos); the backing buffer needs
-        // `ceil(length / 8)` bytes. Keep `length` in bits to stay consistent
-        // with `new_with_values`, `set` and `last_index_of`.
         let bytes = (length as f64 / 8.0).ceil() as usize;
         BitField8 {
             length,
@@ -199,17 +196,11 @@ mod tests {
 
     #[test]
     fn new_uses_bit_length() {
-        // `new(n)` must store the number of *bits* (videos) in `length`, the
-        // same convention used by `new_with_values` and `set`. Storing the byte
-        // count instead makes `last_index_of` (which iterates over `0..length`)
-        // under-scan the bitfield.
         let bf = BitField8::new(20);
         assert_eq!(bf.length, 20);
 
         let mut bf = BitField8::new(20);
         bf.set(15, true);
-        // The last set bit is at index 15, well past the first ceil(20/8) = 3
-        // positions that the buggy length would limit the scan to.
         assert_eq!(bf.last_index_of(true), Some(15));
     }
 }
--- a/stremio-watched-bitfield/src/bitfield8.rs
+++ b/stremio-watched-bitfield/src/bitfield8.rs
@@ -15,10 +15,13 @@ pub struct BitField8 {
 
 impl BitField8 {
     pub fn new(length: usize) -> BitField8 {
-        let length = (length as f64 / 8.0).ceil() as usize;
+        // `length` is the number of bits (videos); the backing buffer needs
+        // `ceil(length / 8)` bytes. Keep `length` in bits to stay consistent
+        // with `new_with_values`, `set` and `last_index_of`.
+        let bytes = (length as f64 / 8.0).ceil() as usize;
         BitField8 {
             length,
-            values: vec![0; length],
+            values: vec![0; bytes],
         }
     }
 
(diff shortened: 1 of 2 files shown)


kKaskak (approved):

tymmesyde:
this PR  doesn't fix anything because it's just renaming variables and adding tests, what is the purpose?

kKaskak:
> this PR doesn't fix anything because it's just renaming variables and adding tests, what is the purpose?

please check the func again, there are two vars including local which causes to shadow the parameter so it changes the meaning

kKaskak:
@tymmesyde 
`BitField8::new(20)` -> requested length 20 -> bytes needed `ceil(20/8) = 3` -> shadow -> length = 3 -> then Bitfield struct becomes `Bitfield8 {length: 3, values: vec![0,3]}`

tymmesyde:
ah okay my bad

kKaskak:
> ah okay my bad

no worries 👍🏼

PR #992, opened 2026-06-06
Player: fix next episode intermittently falling back to the stream picker

Aztup:
## Problem
Clicking "next episode" sometimes plays the next episode directly, but sometimes drops you back to the stream picker and asks you to choose a stream again even when the current stream already covers every episode.

The cause is an update-ordering race in the player model. In the `ResourceRequestResult` handler, `next_video_update` runs **before** `next_stream_update`. So when the next episode's streams finish loading, `next_video.streams` is computed against a stale `next_stream` (`None`) and is left empty. `next_stream_update` then matches the stream, but `next_video` was already built and `next_video_update` does not run again on playback progress, only on a load or another resource result.

That makes the outcome depend on which resource resolves last:
- another resource (e.g. subtitles) resolves **after** the streams → `next_video_update` re-runs → `deepLinks.player` is populated → next episode works.
- the streams resolve **last** → `next_video_update` never re-runs → `deepLinks.player` stays `null` → fallback to the stream picker.

Hence the intermittent behavior.

## Summary
- reorder `next_stream_update` to run **before** `next_video_update` in the `ResourceRequestResult` handler
- the matched next stream is now applied to `next_video.streams` in the same update tick, regardless of resource load order
- `deepLinks.player` is populated deterministically, so "next episode" stops falling back to the stream picker

## Verification
- binge-watched a series: "next episode" now consistently auto-plays instead of intermittently opening the picker

### DIFF
--- a/src/models/player.rs
+++ b/src/models/player.rs
@@ -831,6 +831,9 @@ impl<E: Env + 'static> UpdateWithCtx<E> for Player {
                     _ => Effects::none().unchanged(),
                 };
 
+                let next_stream_effects =
+                    next_stream_update(&mut self.next_stream, &self.next_streams, &self.selected);
+
                 let next_video_effects = next_video_update(
                     &mut self.next_video,
                     &self.next_stream,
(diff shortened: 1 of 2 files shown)


kKaskak (approved):

kKaskak:
@Aztup mind messaging me on discord? my handle is `kkaskak`, i will be testing this thoroughly and after that we will give you a lifetime supporter sub for free. 🎊

Aztup:
Not at all, I’ve sent you a fr

PR #1009, opened 2026-07-24
Player: Ignore invalid outro timestamps

kKaskak:


Botsy (approved):
Tested and issue with outro being 0 is fixed.

PR #1014, opened 2026-07-27
Fix repeated identical player loads not emitting updates

ruizkinio:
## Summary

- treat every `ActionLoad::Player` dispatch as an observable player lifecycle
- preserve equality-based state assignment while forcing the Player field's changed bit
- add a regression proving two identical loads each emit `NewState([Player])`

## Problem

An identical second `ActionLoad::Player` currently makes `eq_update` return unchanged. The derived model then omits `Field::Player` from the runtime event. A client that creates a fresh observer with `getInitialState = false` can miss the load completely and remain in its loading state.

A player load is an action/lifecycle boundary, not only a value assignment, so an identical selection still needs to be observable.

## Implementation

The existing `eq_update` still controls assignment. Joining `Effects::none()` marks the field changed without adding messages or futures.

## Validation

- `cargo fmt --all -- --check`
- `cargo test unit_tests::player:: --lib` (2 passed)
- `cargo test --lib` (193 passed)
- independent review of `Effects::join` semantics and regression scope


kKaskak:
please update branch and tests thx

PR #1015, opened 2026-07-27
Fix calendar today in local timezone

### LINKED ISSUE #1005: Calendar "today" computed in UTC, not local time — highlights the wrong day in evening/morning timezones
### Describe the bug
The Calendar's "today" is computed from the current time in **UTC** and never converted to the user's local timezone. For any user whose local date differs from the UTC date — e.g. the Americas in the evening, or Asia/Oceania in the early morning — the calendar highlights the **wrong day** (one day ahead or behind).

### Root cause
In `src/models/calendar.rs`, the "today" marker is derived directly from `E::now()`:

```rust
let current_date = E::now();
...
let today = if current_date.year() == *year && current_date.month() == *month {
    // uses current_date.day() / weekday() directly
```

`E::now()` returns a UTC timestamp, and `current_date.year()/month()/day()` are read straight off it with no conversion to the user's local timezone. So "today" is always the UTC calendar day, regardless of the client's actual local date.

### To Reproduce
1. Set the system/browser timezone to `America/New_York` (UTC−4/−5).
2. After ~8:00 PM local (i.e. past midnight UTC), open the Calendar.
3. The "today" highlight lands on **tomorrow's** date.

Reproduces identically in the desktop app **and** in `https://web.strem.io` in a normal browser with a correct local timezone — confirming the day is decided in core, independent of any client/JS timezone.

### Expected behavior
"Today" should reflect the user's **local** calendar date, not the UTC date.

### Notes
- `stremio-web`'s `src/routes/Calendar/useCalendarDate.ts` only *formats* labels; the today/selected day comes from this core model, so the fix belongs here.
- Likely needs `E::now()` converted to local time (or a local-date helper) before extracting year/month/day/weekday — depends on whether the `Env` exposes a local-time offset.

### Environment
- stremio-web via `web.strem.io` and the Linux desktop shell (`stremio-linux-shell`, WebKitGTK)
- Timezone: `America/New_York`

kKaskak:
Closes #1005

### DIFF
--- a/src/models/calendar.rs
+++ b/src/models/calendar.rs
@@ -173,7 +173,7 @@ fn selected_update<E: Env + 'static>(
     selected: &mut Option<Selected>,
     next_selected: &Option<Selected>,
 ) -> Effects {
-    let current_date = E::now();
+    let current_date = E::local_now().date_naive();
 
     let updated_selected = next_selected
         .as_ref()
@@ -190,7 +190,7 @@ fn month_info_update<E: Env + 'static>(
     let updated_month_info = selected
         .as_ref()
         .map(|Selected { month, year, .. }| {
-            let current_date = E::now();
+            let current_date = E::local_now().date_naive();
 
             let today = if current_date.year() == *year && current_date.month() == *month {
                 Some(current_date.day())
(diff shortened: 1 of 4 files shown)


Botsy (approved):
lgtm

PR #1031, opened 2026-08-19
feat(streaming-server): add StopOnDevice and SetDeviceSubtitles actions

dioKaratzas:
### Problem

`ActionStreamingServer::PlayOnDevice` starts a casting session, but nothing can stop it or set the subtitles of the remote player. The UI therefore has no counterpart for either:

- leaving the player keeps the device playing, with no way to stop it from the app — Stremio/stremio-bugs#2733
- the selected subtitles track never reaches the device, so subtitles render locally only — Stremio/stremio-bugs#2732

Stremio 4.4 did both, through the same endpoint.

### Changes

The streaming server implements the stremio-cast protocol on `/casting/{device}/player`: a `POST` sets player properties, `source: null` stops the playback, `subtitlesSrc` sets the subtitles URL.

- `ActionStreamingServer::StopOnDevice { device }` → `POST {"source": null}`
- `ActionStreamingServer::SetDeviceSubtitles { device, subtitlesSrc }` → `POST {"subtitlesSrc": ...}`
- `Internal::StreamingServerStopOnDeviceResult` / `StreamingServerSetDeviceSubtitlesResult`
- `Event::StoppedOnDevice { device }`, mirroring `PlayingOnDevice`

`SetDeviceSubtitles` is gated on the device being present in `playback_devices`, like `PlayOnDevice`. `StopOnDevice` deliberately is not: a session must be stoppable even if the device list has not been refreshed since playback started.

Verified against a running streaming server — `GET /casting/{device}/player` on a live session returns `subtitlesSrc`, `subtitlesDelay`, `subtitlesSize` alongside `source`, and devices advertise `"playerUIRoles":[...,"subtitles",...]`.

### Consumer

https://github.com/Stremio/stremio-web/pull/1376 uses these actions (it previously called the endpoint from the web layer, which was correctly flagged in review).

### Note

I have no Rust toolchain on this machine, so this has not been compiled locally — please let CI weigh in; I'll fix anything it flags. `subtitlesDelay` / `subtitlesSize` are left out on purpose: the endpoint accepts them, but their scale differs from the in-app settings and I did not want to guess the mapping.

### DIFF
--- a/src/models/streaming_server.rs
+++ b/src/models/streaming_server.rs
@@ -11,6 +11,7 @@ use crate::models::common::{eq_update, Loadable};
 use crate::models::ctx::{Ctx, CtxError};
 use crate::runtime::msg::{
     Action, ActionStreamingServer, CreateTorrentArgs, Event, Internal, Msg, PlayOnDeviceArgs,
+    SetDeviceSubtitlesArgs, StopOnDeviceArgs,
 };
 use crate::runtime::{Effect, EffectFuture, Effects, Env, EnvError, EnvFutureExt, UpdateWithCtx};
 use crate::types::addon::ResourcePath;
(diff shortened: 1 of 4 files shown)


dioKaratzas:
Closing — the web-side PR this belonged to is being split into three focused PRs per Stremio/stremio-web#1376 review. Will re-open the core part only if the split work still needs it.

PR #1032, opened 2026-08-19
feat: option to enable/disable add-ons without uninstalling

### LINKED ISSUE #1028: Option to enable/disable add-ons without uninstalling
It will be great to have an option to enable disable add-ons.

One use case could be having backup add-ons setup and disabled to avoid streams list polluted/crowded. But if the primary add-on goes down the backup one can be just enabled from add-ons pane.

OussamaERrafif:
Resolves #1028

This PR adds a `disabled` boolean flag to `DescriptorFlags`. When an addon is marked as disabled, it is filtered out from `AggrRequest::plan` and selectable catalogs in `CatalogWithFilters`, meaning it will not be queried for streams, catalogs, subtitles, etc., essentially acting as disabled.

The addons will still show up in the `InstalledAddonsWithFilters` model, allowing users (via front-end UI) to re-enable them by updating their `disabled` flag via the existing `InstallAddon` action.

### DIFF
--- a/src/models/catalog_with_filters.rs
+++ b/src/models/catalog_with_filters.rs
@@ -305,6 +305,7 @@ fn selectable_update<T: CatalogResourceAdapter>(
     let selectable_catalogs = profile
         .addons
         .iter()
+        .filter(|addon| !addon.flags.disabled)
         .flat_map(|addon| {
             T::catalogs(&addon.manifest)
                 .iter()
--- a/src/models/catalogs_with_extra.rs
+++ b/src/models/catalogs_with_extra.rs
@@ -77,6 +77,7 @@ impl<E: Env + 'static> UpdateWithCtx<E> for CatalogsWithExtra {
                         .profile
                         .addons
                         .iter()
+                        .filter(|addon| !addon.flags.disabled)
                         .find(|addon| addon.transport_url == request.base)
                         .and_then(|addon| {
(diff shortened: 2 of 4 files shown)


kKaskak:
hey we added a functionality which is way better than this and allows you to disable per catalog from add-on which solved the issue 
check stremio.com/donate

justauserid:
> hey we added a functionality which is way better than this and allows you to disable per catalog from add-on which solved the issue check stremio.com/donate

I do donate. 

Will wait for the release now and try to understand what "disable per catalog from add-on" means. Thanks.

kKaskak:
> > hey we added a functionality which is way better than this and allows you to disable per catalog from add-on which solved the issue check stremio.com/donate
> 
> I do donate.
> 
> Will wait for the release now and try to understand what "disable per catalog from add-on" means. Thanks.

you have content settings in the settings of the app available 
by disabling all catalogs from addon you essentially disable the addon itself, except the subtitles addons

Screenshot 2026-08-20 at 16 41 09

kKaskak:
these are already live features on all platforms on newest app versions

justauserid:
> these are already live features on all platforms on newest app versions

I get it why I can't see that feature. I donate separately (and the email I use to login to Stremio is nowhere related/connected to the ones from where I pay). I do not subscribe to a Stremio plan. But good to know. I will drop the quest for this feature. Thanks and cheers. (edit: I think there's no donate option any more)

PR #1043, opened 2026-09-03
fix(player): fall back to single next stream without a binge group

Kohafin:
## Problem

Clicking the next-episode popup (or the next button) while playing a stream from an addon that does not set `behaviorHints.bingeGroup` — notably the built-in **Local Files** addon — never auto-plays. `next_stream_update` only accepts a stream whose `bingeGroup` exactly equals the current stream's (`Stream::is_binge_match`), with no fallback. `next_stream` stays `None`, the next video's `deepLinks.player` is `null`, and the web UI falls through to `metaDetailsStreams` — dropping the user on the stream-selection screen for every episode.

Reproduced on Stremio 5.1.26 + local streaming server 4.20.16: `/local-addon/stream/series/<id>.json` returns a single stream with no `behaviorHints`, so binge watching of local files is broken 100% of the time.

## Fix

When no binge-group match is found but the addon returned **exactly one** stream for the next video, use that stream. This is conservative: addons returning multiple streams (where auto-picking could choose badly) keep the existing behavior.

## Testing

- New unit test `next_stream_without_binge_group` covering the single-stream/no-binge-group case.
- `cargo test`: 237 + 18 doc tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/models/player.rs
+++ b/src/models/player.rs
@@ -1156,12 +1156,23 @@ fn next_stream_update(
         ) => streams
             .iter()
             .find(|next_stream| next_stream.is_binge_match(stream))
-            .or_else(|| match streams.as_slice() {
+            .or_else(|| {
                 // Some addons (e.g. Local Files) do not set behaviorHints.bingeGroup;
-                // when the addon returned exactly one stream for the next video,
-                // use it so binge watching still works.
-                [next_stream] => Some(next_stream),
-                _ => None,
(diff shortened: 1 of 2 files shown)


kKaskak:
hello this is to the addon to provide the binge watching group and we cannot implement a fallback here sorry :D most likely due legal, even if its one stream only again, and we have no way to implement to local files the binge watching as well

PR #1045, opened 2026-09-04
feat(subtitles): keep add-on specific subtitle properties

zond:
The addon protocol only specifies `id`, `url` and `lang` for a subtitle object, but add-ons routinely send more. The OpenSubtitles v3 add-on, for instance, attaches `subtitleFileName`, `movieReleaseName`, `releaseGroup` and `fpsMilli` to every entry, and `SubEncoding` is not modelled either. All of it is currently discarded by serde, so a client cannot tell fifteen English uploads apart, cannot label them by release group, and cannot warn that a 25 fps subtitle will drift against a 23.976 fps video.

Rather than blessing one add-on's property names with dedicated fields, collect whatever else was sent into a flattened catch-all, the same way `StreamBehaviorHints::other` and `MetaItemBehaviorHints::other` already do. The properties are serialized back out, so they reach the UI through `stremio-core-web`'s player serializer without any change there.

This is additive: known fields are unaffected, a subtitle without extra properties round-trips to exactly the same JSON as before, and `Default` still works. The only visible difference is that `Subtitles` now (de)serializes as a map instead of a struct, which is what `#[serde(flatten)]` requires and which the serde_test tokens are updated for.

Deliberately limited to `Subtitles`: `Stream` and `MetaItem` drop add-on properties in the same way and deserve the same treatment, but they are larger types with more construction sites and deep-link encoding to think about, so they are better served by follow-ups than by widening this change.

This MR was created by asking my friendly Opus 5 agent to solve this issue.

### DIFF
--- a/src/types/resource/subtitles.rs
+++ b/src/types/resource/subtitles.rs
@@ -1,8 +1,11 @@
+use std::collections::HashMap;
+
 #[cfg(test)]
 use derivative::Derivative;
 use serde::{Deserialize, Serialize};
 use url::Url;
 
+/// See <https://github.com/Stremio/stremio-addon-sdk/blob/master/docs/api/responses/subtitles.md> for documentation
 #[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Debug)]
 #[cfg_attr(test, derive(Derivative))]
 #[cfg_attr(test, derivative(Default))]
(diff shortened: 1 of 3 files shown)


kKaskak:
hey please remove claude as co-author, remove unnecessary big comment blocks, fix lint and builds

zond:
> hey please remove claude as co-author, remove unnecessary big comment blocks, fix lint and builds

Check now, I think this is better.

PR #1046, opened 2026-09-04
Reject non-http streaming server URLs

kKaskak:
`UpdateSettings` accepted any `Url` for `streaming_server_url`, so `?streamingServerUrl=javascript:x` on web made `get_settings` panic on `join` and the wasm worker died until reload.

Non-http(s) URLs are now rejected with an error event and the profile is left untouched. Web side is Stremio/stremio-web#1414.


Botsy (approved):
Tested and all looks good 👍

PR #1047, opened 2026-09-05
Confirm cache folder changes before updating settings

kKaskak:
Keep the accepted cache folder until the server confirms the change, and preserve settings on failure. Send only `cacheRoot` and ignore results after changing servers.

Pass through the server's folder-selection capability for Stremio/stremio-web#1279.


TRtomasz (approved):
tested on windows

